PR #46, opened 2019-10-15
Alignment fix in 'bastille.conf'.

olgeni:
This patch looks perfectly useless, but it's the first thing I noticed after installation 😃


cedwards:
Looks good. I'll include this in the next merge.

PR #52, opened 2019-11-16
Update rule in pf.conf

jpmens:
I'm way out of my depth here, but I followed [these instructions](https://docs.bastillebsd.org/en/latest/chapters/networking.html) yesterday, and PF could be restarted; copy/pasting from Github today, showed this error:

```
Enabling pfno IP address found for lo1:network
/etc/pf.conf:7: could not parse host specification
```

somebody please check this *very* carefully?

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ set block-policy drop
 scrub in on $ext_if all fragment reassemble
 
 set skip on lo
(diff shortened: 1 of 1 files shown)


cedwards:
This fix would work, however I have updated the documentation to clarify this process.

The documentation now instructs users to assign an IP address to the interface, which avoids this error.

PR #66, opened 2019-11-23
minor fixes discovered during 0.5.20191125 RC

cedwards:
Updates to the network bootstrap in the README
added sha256 output validation message in bootstrap
cleaned up some output in zfs


cedwards:
also standardizes on `case` statement for optional `update` command to bootstrap

PR #72, opened 2019-11-26
Additional network config checks

JRGTH:
Added additional network configuration checks


cedwards:
Looks good. Thank you.

PR #85, opened 2019-12-17
Improved container update process, add support for Thickjail updates

JRGTH:
This improve the container/base update process and adds pager, also add support for Thickjail updates too.

### DIFF
--- a/usr/local/share/bastille/update.sh
+++ b/usr/local/share/bastille/update.sh
@@ -32,7 +32,7 @@
 . /usr/local/etc/bastille/bastille.conf
 
 usage() {
-    echo -e "${COLOR_RED}Usage: bastille update release.${COLOR_RESET}"
+    echo -e "${COLOR_RED}Usage: bastille update release | container.${COLOR_RESET}"
     exit 1
 }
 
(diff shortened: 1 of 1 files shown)


usr/local/share/bastille/update.sh
```diff
@@ -47,17 +47,42 @@ if [ $# -gt 1 ] || [ $# -lt 1 ]; then
...
 
 if [ ! -z "$(freebsd-version | grep -i HBSD)" ]; then
     echo -e "${COLOR_RED}Not yet supported on HardenedBSD.${COLOR_RESET}"
     exit 1
 fi
 
-if [ -d "${bastille_releasesdir}/${RELEASE}" ]; then
-    freebsd-update -b "${bastille_releasesdir}/${RELEASE}" fetch install --currently-running "${RELEASE}"
+if [ -d "${bastille_jailsdir}/${TARGET}" ]; then 
+    if ! cat "${bastille_jailsdir}/${TARGET}/fstab" 2>/dev/null | grep -w "${TARGET}" | grep -q ".bastille"; then
```
cedwards: This can be simplified into something like: `grep -w "${TARGET}" "${bastille_jailsdir}/${TARGET}/fstab" | grep -q ".bastille"`

usr/local/share/bastille/update.sh
```diff
@@ -47,17 +47,42 @@ if [ $# -gt 1 ] || [ $# -lt 1 ]; then
...
+            if [ "$(jls name | grep -w "${TARGET}")" ]; then
+                # Update a thick container.
+                CURRENT_VERSION=$(jexec -l ${TARGET} freebsd-version 2>/dev/null)
```
cedwards: Let's standardize on using the full path to binaries called using `jexec`, ie; `/bin/freebsd-version`.

cedwards:
I like the ability here to update releases or standalone containers. Please have a look at my comments and we'll get merged.

JRGTH:
> I like the ability here to update releases or standalone containers. Please have a look at my comments and we'll get merged.

Hi, I will simplify the code and update paths as requested, so will update the PR during the morning after testing and confirm working as expected.

Regards

usr/local/share/bastille/update.sh
JRGTH: Hi, we can grep directly on the target/fstab which is even simpler:
`grep -qw ".bastille" "${bastille_jailsdir}/${TARGET}/fstab"`

I will test and update PR today

PR #86, opened 2019-12-23
support for ressource limits; small Makefile improvemts

hackacad:
running `make` (without args) automatically used case `install`.


hackacad:
added ressource limits
small Makefile fix

PR #88, opened 2019-12-27
Fix to match exact container name, add missing quotes

JRGTH:
This fix for matching exact container name and deals with names containing multiples dash/hyphen, also added missing quotes to avoid "unexpected operator" errors.


cedwards:
Thank you.

PR #89, opened 2019-12-27
Minor changes, fix notify if container not started

JRGTH:
Hi, this minor change fix for notify if the container is not started yet, was failing previously since placed in the for loop that will do noting if ${JAILS} empty.

Regards!

### DIFF
--- a/usr/local/share/bastille/create.sh
+++ b/usr/local/share/bastille/create.sh
@@ -37,7 +37,13 @@ usage() {
 }
 
 running_jail() {
-    jls name | grep -w "${NAME}"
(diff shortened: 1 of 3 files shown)


JRGTH:
Hi, I've added some jail validation improvements and dealing with dashes on jail names for create and destroy commands.

This also fixes accidental double jail creation, especially for Thick jails where ".bastille" mountpoint/directory may not exist since not needed.

Also fixes the destroy issue on jails with similar names using dashes.

Regards

cedwards:
Looks good, as usual.

Thank you.

PR #94, opened 2020-01-15
build: fix manual install path

dch:
I'm assuming that's what you expected here, could equally have been something in `/usr/local/share`.


dch:
- `/usr/local/bastille` doesn't exist by default so it you *first* run bastille from git, so create it

cedwards:
Thank you. Good catch.

PR #95, opened 2020-01-15
Print a simple directory tree to verify a template's contents using f…

bsdlme:
…ind + sed if tree(1) is not installed


cedwards:
Nice alternative. Thank you.

PR #96, opened 2020-01-15
initial man page

hackacad:
will be adding all options as the bastille feature set (options/args) will become more clear


cedwards:
This PR is the story of how I created an issue to remind myself and the community closed the bug in under a week's time. Thank you.

PR #97, opened 2020-01-16
Add support for thinjail to thickjail conversion, minor code fixes

JRGTH:
Hi, this add support for Thinjail to Thickjail conversions and some minor code fixes.

Usage:
 `bastille convert name`

Regards

### DIFF
--- a/usr/local/share/bastille/convert.sh
+++ b/usr/local/share/bastille/convert.sh
@@ -1,0 +1,168 @@
+#!/bin/sh
+# 
+# Copyright (c) 2018-2020, Christer Edwards <christer.edwards@gmail.com>
+# All rights reserved.
+# 
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+# 
+# * Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+# 
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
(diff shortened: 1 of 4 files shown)


JRGTH:
Hi, added support to re-bootstrap additional distfiles after updating bootstrap_archives.

Solves [unable to re-bootstrap after updating bootstrap_archives in config #98](https://github.com/BastilleBSD/bastille/issues/98)

Regards

usr/local/share/bastille/convert.sh
```diff
@@ -0,0 +1,168 @@
...
+    error_notify "${COLOR_GREEN}Changes for '${TARGET}' has been reverted.${COLOR_RESET}"
+}
+
+start_convert() {
+    # Attempt container conversion and handle some errors
+    if [ -d "${bastille_jailsdir}/${TARGET}" ]; then
+        echo -e "${COLOR_GREEN}Converting '${TARGET}' into a thickjail, this may take a while...${COLOR_RESET}"
+
+        # Set some variables
+        RELEASE=$(grep -owE '([1-9]{2,2})\.[0-9](-RELEASE|-RC[1-2])' ${bastille_jailsdir}/${TARGET}/fstab)
```
cedwards: Can we improve this by detecting which version the container is running (ie; `jexec -l ${TARGET} /bin/freebsd-version | sed 's/-p.*//'`) and slot that in for the `fstab` search? This way we're not casting a regex net, but specifically matching the only version it should be.

cedwards:
This looks good and nice addition.

If you're ever up for a challenge I think it's possible to do convert in the opposite direction, just a little harder.

Issue #101, opened 2020-01-23
bastille_zfs_mountpoint is GNDN


backerman:
`bastille.conf` contains a `bastille_zfs_mountpoint` option to change where the jails are mounted, but it's not actually used anywhere in the code and therefore has no effect.

cedwards:
This has been resolved.

`bastille_zfs_mountpoint` removed; mounts to `bastille_prefix`.

PR #102, opened 2020-01-25
bug #101 bastille.conf fix

bmac2:


cedwards:
I think the preferred solution here is to implement `${bastille_zfs_mountpoint}` vs remove it completely.

PR #103, opened 2020-01-26
make escaped quote visible

jpmens:


cedwards:
Thank you for the attention to detail.

Issue #123, opened 2020-02-03
[ENHANCEMENT] check if IP address exists


hackacad:
Implement a check if you created ip address exists.
Otherwise you will run into errors like:

`jail: jailname: /sbin/ifconfig bastille0 inet 10.0.0.1 netmask 255.255.255.255 alias: failed`

Could be done by a simple grep on ifconfig (ifconfig | grep ipadress), but maybe there's a more reliable solution

hackacad:
close #138

PR #124, opened 2020-02-03
Check if IP address is in use

hackacad:
corresponding issue: #123 

This will check if an IP address is already in use.
It won't work if a jail (with that IP address) is not running, which should be fine and maybe wanted in some cases (reinstall of existing jail).

I'll add the same feature for "bastille start" as well.


hackacad:
Dropped the idea to fail on "bastille create". There's might be cases where you want to create a "successor" with the same IP address. 
See #125

PR #125, opened 2020-02-03
Check if IP address is in use

hackacad:


hackacad:
#125 changes:
(1) warning when running create.sh
(2) error and exit when starting

PR #126, opened 2020-02-03
Templates for bug reports.

hackacad:
adding templates for bug reports

### DIFF
--- a/usr/local/share/bastille/create.sh
+++ b/usr/local/share/bastille/create.sh
@@ -58,7 +58,11 @@ validate_ip() {
           exit 1
         fi
       done
-      echo -e "${COLOR_GREEN}Valid: ($ip).${COLOR_RESET}"
+      if ifconfig | grep inet | grep "$ip" >/dev/null; then
+        echo -e "${COLOR_YELLOW}Warning: ip address already in use ($ip).${COLOR_RESET}"
+      else
+        echo -e "${COLOR_GREEN}Valid: ($ip).${COLOR_RESET}"
+      fi
     else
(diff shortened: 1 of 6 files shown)


usr/local/share/bastille/create.sh
```diff
@@ -58,7 +58,11 @@ validate_ip() {
           exit 1
         fi
       done
-      echo -e "${COLOR_GREEN}Valid: ($ip).${COLOR_RESET}"
+      if ifconfig | grep inet | grep -w "$ip" >/dev/null; then
```
cedwards: Can we just `grep -w "$ip"` without the second grep?

usr/local/share/bastille/start.sh
```diff
@@ -64,6 +64,12 @@ for _jail in ${JAILS}; do
 
     ## test if not running
     elif [ ! "$(jls name | awk "/^${_jail}$/")" ]; then
+	ip=$(cat ${bastille_jailsdir}/${_jail}/jail.conf | grep ip4.addr | awk '{print $3}' | sed 's/\;//g')
```
cedwards: `$(grep 'ip4.addr' "${bastille_jailsdir}/${_jail}/jail.conf" | awk '{print $3}' | sed 's/\;//g')`

usr/local/share/bastille/create.sh
hackacad: Not sure if there's a case where the "IP address" is listed in any other line you don't want.
But I think it should be fine to remove the inet grep

PR #133, opened 2020-02-03
Readme update

hackacad:
some improvements to rdr

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ Create the firewall config, or merge as necessary.
 ------------
 ```
 ext_if="vtnet0"
+rdr_tcp = "{80, 443}" # bastille TCP ports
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -21,7 +21,8 @@ portsnap fetch auto
 make -C /usr/ports/sysutils/bastille install clean
 ```
 
-**Git**
+**Git (for testing only)**
+This will break your running (config) files
```
cedwards: Is this message still valid if we merge the `.sample` PR?

hackacad: This is plan B if we don’t switch to .sample. 
Might be closed

hackacad: This is plan B if we don’t switch to .sample. 
Might be closed

PR #135, opened 2020-02-03
change bastille.conf to bastille.conf.sample

hackacad:
prevents overwriting config file while updating as suggested by @cedwards


usr/local/etc/bastille/bastille.conf.sample
```diff
@@ -28,8 +28,8 @@ bastille_url_freebsd="http://ftp.freebsd.org/pub/FreeBSD/releases/"          ##
 bastille_url_hardenedbsd="http://installer.hardenedbsd.org/pub/hardenedbsd/" ## default: "https://installer.hardenedbsd.org/pub/HardenedBSD/releases/"
 
 ## ZFS options
-bastille_zfs_enable=""                                                ## default: ""
-bastille_zfs_zpool=""                                                 ## default: ""
+bastille_zfs_enable="NO"                                              ## default: "NO"
```
cedwards: Is there a reason we need to reset these default values?

PR #139, opened 2020-02-04
Compare/update fstab if both variables are set

JRGTH:
Hi this simplifies the fstab content comparison on thin containers.
This fix for the `sed: first RE may not be empty` error during remote container imports.

Regards

### DIFF
--- a/usr/local/share/bastille/import.sh
+++ b/usr/local/share/bastille/import.sh
@@ -95,21 +95,17 @@ update_jailconf() {
 
 update_fstab() {
     # Update fstab .bastille mountpoint on thin containers only
-    # Check if is a thin container
-    if [ ! -d "${bastille_jailsdir}/${TARGET_TRIM}/root/.bastille" ]; then
-        break
-    elif ! grep -qw ".bastille" "${bastille_jailsdir}/${TARGET_TRIM}/fstab"; then
-        break
-    fi
-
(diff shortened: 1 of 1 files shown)


usr/local/share/bastille/import.sh
```diff
@@ -95,21 +95,17 @@ update_jailconf() {
...
-    if [ ! -d "${bastille_jailsdir}/${TARGET_TRIM}/root/.bastille" ]; then
-        break
-    elif ! grep -qw ".bastille" "${bastille_jailsdir}/${TARGET_TRIM}/fstab"; then
-        break
-    fi
-
     # Set some variables
     FSTAB_CONFIG="${bastille_jailsdir}/${TARGET_TRIM}/fstab"
     FSTAB_RELEASE=$(grep -owE '([1-9]{2,2})\.[0-9](-RELEASE|-RC[1-2]|-stable-build-[0-9]{1,3})' ${FSTAB_CONFIG})
     FSTAB_CURRENT=$(cat ${FSTAB_CONFIG} | grep -w ".*/releases/.*/jails/${TARGET_TRIM}/root/.bastille")
```
cedwards: unnecessary `cat`.

JRGTH: Right, will check and update in a bit.

PR #140, opened 2020-02-04
improve bastille list

hackacad:
This patch should fix the following issues:

(1) `bastille list`: has been using jls -N
- what happened?: showed bastille jails as well as manually created or from other jail managers
- new behavior: show list of arg2 (help function)

(2) `bastille list jail|jails|container|containers`:  greped all jail names from ${bastille_jailsdir}/${_JAIL}/jail.conf
- new behavior:  lists all bastille jail incl. IP-address, path and status

### DIFF
--- a/usr/local/share/bastille/list.sh
+++ b/usr/local/share/bastille/list.sh
@@ -37,7 +37,8 @@ usage() {
 }
 
 if [ $# -eq 0 ]; then
-   jls -N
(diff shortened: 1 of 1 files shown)


hackacad:
might be changed to table view (e.g. `column -t -s' ' `)

cedwards:
Changes to the list output have wider implications. This breaks the start command.

Let's discuss UI changes first. Rejecting for now; list changes can be proposed for 0.7.x after wider testing.

PR #152, opened 2020-02-14
Ignore IPv4 check if there is no entry at all

JRGTH:
This will Ignore IPv4 checking if there is no related `ip4.addr` entry at all, i.e. for IPv6 only containers.

### DIFF
--- a/usr/local/share/bastille/create.sh
+++ b/usr/local/share/bastille/create.sh
@@ -122,10 +122,10 @@ create_jail() {
     if [ ! -d "${bastille_jailsdir}/${NAME}" ]; then
         if [ "${bastille_zfs_enable}" = "YES" ]; then
             if [ ! -z "${bastille_zfs_zpool}" ]; then
-                ## create required zfs datasets
+                ## create required zfs datasets, mountpoint inherited from system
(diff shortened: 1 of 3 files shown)


JRGTH:
In regards to PR [Don't set jail ZFS dataset mountpoint, let be inherited from the system](https://github.com/BastilleBSD/bastille/pull/152/commits/015558c4bc2e05f53be0ab14f51605240cb773ae)

This will let the newly created jails ZFS dataset mountpoint to be inherited from the system, in favor to simplify the import between systems/managers, and ease the manual administration with ZFS send/recv.

This sill not change the Bastille import checks performed during container imports and ZFS mountpoints will be updated accordingly whenever required as usual.

**Before:**
```
root@server: ~# zfs get mountpoint storage/bastille/jails/plex
NAME                         PROPERTY    VALUE                             SOURCE
storage/bastille/jails/plex  mountpoint  /mnt/storage/bastille/jails/plex  local
```

**After:**
```
root@server: ~# zfs get mountpoint storage/bastille/jails/Plex
NAME                         PROPERTY    VALUE                             SOURCE
storage/bastille/jails/Plex  mountpoint  /mnt/storage/bastille/jails/Plex  inherited from storage/bastille/jails
```

**Note:** This only affects new containers and does not interfere with the existing ones.
Regards

PR #188, opened 2020-04-08
fix zfs snapshot recursive

hackacad:
Snapshots should be created recursively to add "data dataset" (jailname/root) as well


cedwards:
Thank you!

PR #202, opened 2020-04-14
Correct capital E in zonepath to UTC

mariusvw:

PR #203, opened 2020-04-14
Remove etc from UTC in zoneinfo path

mariusvw:


mariusvw:
This is the second solution of two. See #202 for more info.

cedwards:
Thank you for providing two options for this solution. Having merged the other I'll close this.

PR #209, opened 2020-04-21
Add vagrant support for testing

mariusvw:


mariusvw:
In reaction to your question if this file can be lower case.

I think it is smart to keep it this way, other projects use the same casing:
```
Makefile
Dockerfile
etc..
```

PR #215, opened 2020-05-23
Option to force destroy jail in ZFS

JRGTH:


JRGTH:
Small update on PR: [932f1af](https://github.com/BastilleBSD/bastille/pull/215/commits/932f1afae1d7d1d2e348b255a25bf604d4688698)

This PR will let `bastille` to append the actual system environment over the defined environment `${PATH}`.

This will also fix the colors.pre `NULL` variables causing colors not showing on version and error output.

**Current behavior:**
image

**Patch applied:**
image

The colors issue was introduced since the `colors.pre.sh` was removed from script top execution, since most of the code calling for this file are mostly errors, I created a small function for it to be called accordingly and prevent for execution duplicate.

Regards

Issue #93, opened 2020-01-13
Template SERVICE


brumgeek:
I have been creating some test templates, one is a node.js app if I do npm insall -g appname in CMD, and then add it as a service it will start fine from restart of container or console.

If I add SERVICE and the app start it can't start as CMD runs after SERVICE.

Would it be better to change the order of template creation so SERVICE runs last?

cedwards:
I have been thinking about options to define a custom order for template execution. No code yet, but I have some ideas.

In your use-case is it simpler to remove (skip) SERVICE and start the service in CMD as you've done?

brumgeek:
Skip would work, what I had to do was set up the jail, then create the service by hand.
Allowing the user to define a custom order for template execution would be better for this use case as I could have CMD run before SERVICE.

Maybe you could have another file in the template that definies custom execution order?
If that file wasn't in the template it would just run the template in the normal execution order.

cedwards:
Please have a look at the latest template syntax particularly the `Bastillefile` which now allows any-order scripting.

chriswells0:
I'm surely biased toward the `Bastillefile`, but it's worth noting that you should also be able to change the execution order now that #218 is merged:

`HOOKS='LIMITS INCLUDE PRE FSTAB PF PKG OVERLAY CONFIG SYSRC CMD SERVICE' bastille template folsom username/base`

I didn't actually test that, but #218 uses the existing value of `HOOKS` if it's set and the default order otherwise.

PR #217, opened 2020-05-24
Add Bastillefile support to templates.

chriswells0:
Provides the ability to replace all of the "hook" files with a single Bastillefile similar to a Dockerfile.  Each line in the Bastillefile maps to a Bastille command and executes the existing code for the commands.

I aliased `INCLUDE` to `TEMPLATE`, so either can be used in a Bastillefile. I was on the fence about doing that, but I left it in for your feedback. On one hand, it may be a slippery slope to provide multiple commands for the same action. It could snowball into others like `COPY` for `CP`, etc. On the other hand, `INCLUDE` simply makes more sense within the context of a template.

Addresses #93 by allowing commands in any sequence.

Enables support for `RDR` (maybe others too?) without having to create a hook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,6 +696,31 @@ The above example will include anything under "etc" and "usr" inside
 the template. You do not need to list individual files. Just include the
 top-level directory name.
 
+For more control over the order of operations when applying a template,
+create a `Bastillefile` inside the base template directory. Each line in
+the file should begin with an uppercase reference to a Bastille command
+followed by its arguments (omitting the target, which is deduced from the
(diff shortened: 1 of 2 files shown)


chriswells0:
I forgot to mention that I aliased `INCLUDE` to `TEMPLATE`, so either can be used in a Bastillefile.  I was on the fence about doing that, but I left it in for your feedback.  On one hand, it may be a slippery slope to provide multiple commands for the same action.  It could snowball into others like `COPY` for `CP`, etc.  On the other hand, `INCLUDE` simply makes more sense within the context of a template.

PR #231, opened 2020-07-27
Fixed the overlay command in Bastillefile

tobiastom:
The documentation did mention that `OVERLAY` should be used, but only `cp` was implemented.
Also the destination part where to copy the files to has been missing. 

The previous hooks seem to implement it in a similar way: https://github.com/BastilleBSD/bastille/blob/master/usr/local/share/bastille/template.sh#L175

### DIFF
--- a/usr/local/share/bastille/template.sh
+++ b/usr/local/share/bastille/template.sh
@@ -132,7 +132,8 @@ for _jail in ${JAILS}; do
                     # Allow redirection within the jail. -- cwells
                     _args="sh -c '${_args}'"
                     ;;
-                cp)
+                overlay)
+                    _cmd='cp'
                     # Convert relative "from" path into absolute path inside the template directory. -- cwells
                     if [ "${_args%${_args#?}}" != '/' ]; then
                         _args="${bastille_template}/${_args}"
--- a/usr/local/share/bastille/template.sh
+++ b/usr/local/share/bastille/template.sh
@@ -134,6 +134,9 @@ for _jail in ${JAILS}; do
                     ;;
                 overlay)
                     _cmd='cp'
+                    _args="${bastille_template}/${_args} /"
+                    ;;
+                cp|copy)
                     # Convert relative "from" path into absolute path inside the template directory. -- cwells
                     if [ "${_args%${_args#?}}" != '/' ]; then
                         _args="${bastille_template}/${_args}"


chriswells0:
I'm the one responsible for creating these bugs.

The intention was to create a `CP` alias for `OVERLAY`, but I forgot to use `cp|overlay)`. I think it would be nice to have that as an option instead of just changing it to overlay. `COPY` is easier to read, but I used `CP` for consistency with the existing command. I don't know if @cedwards would want to weigh in on those points (both having an alias in general and whether to use `CP` or `COPY`).

As for the missing slash, I was providing a destination path in my testing:

`CP www/ usr/local/www/nginx-dist/`

I see why supporting no 2nd parameter for backward compatibility makes sense, but it would be great to also support a 2nd parameter to have the ability to specify a destination. It's kind of a pain to implement taking different possibilities into consideration: 1 param or 2, either/both of them quoted with spaces, either/both having a leading slash. This function seems to work, but please double check me:

```
make_paths_absolute() {
    _absolute_paths=''

    # First parameter: "from" path.
    if [ "${1%${1#?}}" = '/' ]; then # Absolute path, so just ensure it's quoted. -- cwells
        _absolute_paths="\"${1}\""
    else # Convert relative "from" path into absolute path inside the template directory. -- cwells
        _absolute_paths="\"${bastille_template}/${1}\""
    fi

    # Second parameter: "to" path.
    if [ -z "$2" ]; then # Not provided, so assume root. -- cwells
        _absolute_paths="${_absolute_paths} /"
    elif [ "${2%${2#?}}" = '/' ]; then # Absolute path, so just ensure it's quoted. -- cwells
        _absolute_paths="${_absolute_paths} \"${2}\""
    else # Convert relative "to" path into absolute path inside the jail. -- cwells
        _absolute_paths="${_absolute_paths} \"/${2}\""
    fi

    echo "$_absolute_paths"
}
```

Then you have to call it using eval to parse the single string containing both paths into 2 strings:
```
cp|overlay)
	_cmd='cp'
	_args=$(eval "make_paths_absolute $_args")
	;;
```

Would you mind using that or something similar to get both working?

tobiastom:
Would it help to merge this if I implement it like you described above?

cedwards:
It doesn't need to be implemented like above but we should support both `cp` and `overlay` in the template cases. `cp` provides both source and dest args, while `overlay` provides the source path and assumes `/`(root) of container.

tobiastom:
@chriswells0 Would you mind looking at this again. I fixed it very basic right now for two reasons:

1. I like the separation between `overlay` and `copy`. Not sure if everybody will share this opinion
2. even if I like your implementation of `make_paths_absolute`, I'm not sure if it's trying to be to smart.

With this one copy needs two arguments, and overlay one. 

Also, I added an alias `copy` for `cp`.

cedwards:
I think this looks pretty good. "Basic" is (usually) the goal.

chriswells0:
Sorry for the delay: going through some busy times.

Thanks for keeping the 2-parameter version of `CP`/`COPY`. I like the simplicity of knowing I can use any Bastille subcommand in a Bastillefile.

PR #276, opened 2020-11-26
Display message about file/directory being rendered.

chriswells0:
As requested in:

https://github.com/BastilleBSD/bastille/pull/255#issuecomment-721084808


tobiastom:
Thank you for the improvement! Works for me. 

I would prefer to display every file from a directory that is rendered. This would make the render more explicit, and behaves the same way `CMD` does.

chriswells0:
Updated to display the path of each file in the directory being rendered.

tobiastom:
Works for me.

PR #279, opened 2020-11-26
Use templates for configuring new jails.

### LINKED ISSUE #205: [ENHANCEMENT] Default template
I need to put some extra files on the system, not only /etc/resolve.conf but also my own repo config file like  /usr/local/etc/pkg/repos/poudriere.conf And with that a ssl certificate for poudriere /usr/local/etc/ssl/private and a standard .vim .profile and my ssh authrized_key file, and so on.

A default template that can be filled with personal config files.
I set it now in my bastille.conf file, a nice adition would be a command line toggle  like -t to include the template.
root@BSD # bastille create -t  jailname 12.1-RELEASE 10.10.9.1 

None

**Additional context**
I have made a POC for this and it works for me.

in /usr/local/etc/bastille/bastille.conf i have added the following lines and removed the resolv.conf part because i use my template for this.
```
bastille_default_template="YES"                                       ## default: ""
bastille_default_template_files="${bastille_templatesdir}/default/base" ## default:  
 "${bastille_templatesdir}/default/base"
```
and removed 
```
## default jail resolv.conf
bastille_resolv_conf="/etc/resolv.conf"                               ## default: "/etc/resolv.conf"
```
In /usr/local/share/bastille/create.sh i removed

```
## resolv.conf (default: copy from host)
    if [ ! -f "${bastille_jail_resolv_conf}" ]; then
        cp -L "${bastille_resolv_conf}" "${bastille_jail_resolv_conf}"
```
And replaced it with 
```
     ## Copy files from default template
     if [ ${bastille_default_template+x} ]; then
         echo -e "${COLOR_GREEN}Copy files from default template. ${COLOR_RESET}"
         while read _dir; do
             cp -av "${bastille_default_template_files}/${_dir}" "${bastille_jail_path}" || exit 1
         done < "${bastille_default_template_files}/OVERLAY"
         echo -e "${COLOR_GREEN}[${_jail}]:OVERLAY -- END${COLOR_RESET}"
         echo
```
I also create the default template named base in /usr/local/bastille/templates/default/base
There i have my OVERLAY file and all the files i need on a fresh jail.

For me this works very nice

chriswells0:
I'm providing a draft version of this code for early review.

#287 needs to be merged first. If you're testing this change before it gets merged, be sure to apply that PR first.

Use templates for configuring new jails.
Allow user to override default templates.

Closes #205.


usr/local/share/bastille/bootstrap.sh
```diff
@@ -164,6 +164,7 @@ bootstrap_directories() {
         else
             mkdir -p "${bastille_templatesdir}"
         fi
+        ln -s "${bastille_sharedir}/templates/default" "${bastille_templatesdir}/default"
```
hackacad, line 167: This else will only be hit on new deployments. There needs to be an additional check for existing deployments, otherwise you'll get an error when creating new jails.

```
[thin1]:
thin1: created

default/thick not found.
[thin1]:
thin1: removed
```

usr/local/share/bastille/bootstrap.sh
```diff
@@ -304,6 +305,7 @@ bootstrap_template() {
...
```
hackacad, line 308: same as line 167

hackacad (changes requested):
additional checks need to be implemented for existing bootstraps

usr/local/share/bastille/templates/default/thin/Bastillefile
```diff
@@ -0,0 +1,12 @@
+ARG HOST_RESOLV_CONF=/etc/resolv.conf
+ARG IP
```
tobiastom, line 2: This is not needed right not, right?

usr/local/share/bastille/templates/default/thick/Bastillefile
```diff
@@ -0,0 +1 @@
+INCLUDE default/thin
```
tobiastom, line 1: This could reuse the `bastille_template_thin` variable. With that the thin template could also be configured.

usr/local/share/bastille/bootstrap.sh
chriswells0: Nice catch. I added a check in `create.sh` for this.

usr/local/share/bastille/templates/default/thin/Bastillefile
chriswells0: We do use `HOST_RESOLV_CONF` (though I had a typo in the script, so it wasn't getting passed).

By specifying them, I was trying to address 2 situations:

1. Make it clear what args will be passed to the template in case the user creates their own templates.
2. Consider the args we may need in the future so they're "known" even before we start to use them.

For example, I just added `JAIL_NAME` because we have it available to us and I can see it being needed by a user-defined template.

A more robust approach would be to create config vars for the template args, but that seems like over-engineering. Also, since `create.sh` applies the template, it wouldn't know where to get the values from.

usr/local/share/bastille/templates/default/thick/Bastillefile
chriswells0: We don't have access to that variable inside the template, but I refactored it to have a `default/base` template to address this. Now a user can override/extend the base, which affects thin, thick, and vnet jails by default, or override/extend any of the specific templates for those jail types (including changing whether or not they extend `default/base`).

chriswells0: I just tried to implement my own template and got your point. ;) I've made an update to pass the base template as an arg, so now you can set `bastille_template_base` to a custom template and it will be included by the thick, thin, and vnet templates without having to modify them. In my case, I created a base template that extends `default/base`, so it's as if I wedged changes in between.

PR #287, opened 2020-11-29
Provide JAIL_NAME and JAIL_IP as default template args.

chriswells0:
Made both `${JAIL_NAME}` and `${JAIL_IP}` available in templates without having to define them as args.


tobiastom:
Works great for me.

tobiastom:
Is there anything I can help to get this one, #288 and #279 merged into master?

PR #288, opened 2020-11-29
Properly escape config property and value.

chriswells0:
This resolves an issue setting values that contain a forward slash. I also fixed a potential issue where a dot in the property name could cause a false positive match on a property with a different character in the same position, but that shouldn't ever happen since real property names are not that similar.

You may want to wait for more feedback on the config feature before merging this one since I could amend the PR to include additional changes. I don't mind submitting another PR later if you want to get this fix in, though.


tobiastom:
I can confirm that I can successfully set keys and values with special characters.

PR #305, opened 2021-01-05
Thick Upgrade

hackacad:
How to upgrade VNET jail

obsoletes #274


cedwards:
Thank you for this. Apologies for the delay.

PR #310, opened 2021-01-07
Jail cpu an memory statistics

yerrysherry:
Sometimes you want to know what a jail is doing. The command "bastille stat" gives a sorted jails overview with cpu and memory statistics. It gives an indication. The figures are not 100% correct because libraries are shared between the jails. The cpu% can grow beyond the 100%. If you have 8 cores then the total of the cpu % can grow until 800%. But when you have a high cpu load it is a fast way to find a potential bottleneck.

root@bastille:~ # bastille stat

Hostname                                    %mem  %cpu    rss MB
bugs_bunny                                    1.6         63      15.10
chicken_farm                                  1.6         33      15.11
test                                                   0.6        0        5.27
test1                                                 0.6        0        5.25
this_is_a_very_long_hostname       2           0        19.38

rss = the real memory (resident set) size in MB

If it is interesting, than the command should be tested on different servers. Maybe there is a better solution with dtrace but that's another level. It would be great to also find netwerk and disk IO statistics.

### DIFF
--- a/jls2.sh
+++ b/jls2.sh
@@ -1,0 +1,13 @@
+#/bin/sh
+
+printf "   %-4s %-15s %-28s  %-36s  %-10s  %-12s  %-10s\n"       "JID"  "IP Address" "Hostname" "Path" "Memory" "Virtual Mem" "Disk"
+for JID in `jls | grep -v JID | awk '{print $1}'`
+do
+a=`jls -j $JID | sed 1d`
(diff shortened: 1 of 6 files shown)


bmac2:
This PR is on a really outdated codebase.  IF it is still a valid request, please submit a PR from the current code base to be reviewed.

PR #311, opened 2021-01-08
Don't start empty jails, don't apply template if Bastillefile is empty

JRGTH:
Hi, this PR will solve the empty jail creation errors, since `bastille` is trying to start/restart and apply default template to it, but it is pretty much a nulljail without content initially as expected.

**Error when creating Empty jails:**
```
root@nas-mserver: ~# bastille create -E nulljail
Creating empty jail: nulljail.
grep: Unmatched [ or [^
Error: not set interface does not exist.
[nulljail]: Not started. See 'bastille start nulljail'.
[nulljail]: Not started. See 'bastille start nulljail'.
grep: Unmatched [ or [^
Error: not set interface does not exist.
```

Since empty jails are intended for experimental and/or unsupported jail setups, this template apply function most likely will be unused, though said code will be there for reference and/or future use.

**Output after simple correction:**
```
root@nas-mserver: ~# bastille create -E nulljail
Creating empty jail: nulljail.
```

The simple fix I've propose is to just ignore the template apply for the empty jail if the `Bastillefile` is empty which makes sense, though this may change in the future.

Regards


yerrysherry (commented):
Hostname can not start or end with -. Also in the middle of the name two dashes are not allowed.

Those names are not valid:
-test
test-
test--test

Maybe using this as a starting point:

echo test | grep -E '^([a-z0-9])+(-[a-z0-9]+)*([a-z0-9a])*$'
test
echo test- | grep -E '^([a-z0-9])+(-[a-z0-9]+)*([a-z0-9a])*$'
echo -test- | grep -E '^([a-z0-9])+(-[a-z0-9]+)*([a-z0-9a])*$'
echo test--test | grep -E '^([a-z0-9])+(-[a-z0-9]+)*([a-z0-9a])*$'
echo test-test | grep -E '^([a-z0-9])+(-[a-z0-9]+)*([a-z0-9a])*$'
test-test

Converting capital letters to lowercase?

yerrysherry:
When copy and paste  '*' was lost

grep -E '^([a-z0-9])+(-[a-z0-9]+)*([a-z0-9a])\*$'

JRGTH:
Hi yerrysherry, you can use `tr` with the upper|lower classes to change capital letter to lower case, here's an example:

```
# echo test-test | grep -E '^([a-zA-Z0-9])+(-[a-zA-Z0-9]+)*([a-zA-Z0-9a])*$' | tr '[:upper:]' '[:lower:]'
test-test
# echo TEST-tEsT | grep -E '^([a-zA-Z0-9])+(-[a-zA-Z0-9]+)*([a-zA-Z0-9a])*$' | tr '[:upper:]' '[:lower:]'
test-test
```

P.S. I think this Hostname naming restrictions should be discussed too, also `bastille create|rename` will set/edit the `host.hostname` same as the jail name initially, but the user can run `bastille edit jail_name` to edit `jail.conf` accordingly as a workaround.

Regards

yerrysherry:
The hostname is for me a valid DNS name without the domain. Indeed, and validating a jail.conf is very hard to do. And not forget, each new FreeBSD has some changes.

hackacad (approved):
Works as expected 👍🏻

usr/local/share/bastille/create.sh
```diff
@@ -174,7 +176,7 @@ ${NAME} {
 
   vnet;
   vnet.interface = e0b_${uniq_epair};
-  exec.prestart += "jib addm ${uniq_epair} ${INTERFACE}";
+  exec.prestart += "jib addm ${uniq_epair} ${bastille_jail_conf_interface}";
```
cedwards, line 179: unsure about this change

usr/local/share/bastille/create.sh
```diff
@@ -261,6 +263,13 @@ create_jail() {
             for _link in ${LINK_LIST}; do
                 ln -sf /.bastille/${_link} ${_link}
             done
+            # Properly link shared ports on thin jails in read-write.
+            if [ -d "${bastille_releasesdir}/${RELEASE}/usr/ports" ]; then
+                if [ ! -d "${bastille_jail_path}/usr/ports" ]; then
+                    mkdir ${bastille_jail_path}/usr/ports
+                fi
+                echo -e "${bastille_releasesdir}/${RELEASE}/usr/ports ${bastille_jail_path}/usr/ports nullfs rw 0 0" >> "${bastille_jail_fstab}"
```
cedwards, line 271: don't know if we need to make this a default. is easily templated however.

usr/local/share/bastille/create.sh
JRGTH: Hi cedwards, this is just the same way how the current `non-vnet` jail rc.conf is applied, i.e. if the user don't have a defined NIC in the `bastille.conf` then the vnet jail always need a NIC defined when creation, if the user had said NIC in the config then, the vnet jail will take that NIC if the user don't specify one in the creation command, however if the user do define a NIC in the vnet jail creation command this NIC in the config is overridden by.

Here is an example of the current `jail.conf` snippet on non-vnet jails:
```
generate_jail_conf() {
    cat << EOF > "${bastille_jail_conf}"
${NAME} {
  devfs_ruleset = 4;
  enforce_statfs = 2;
  exec.clean;
  exec.consolelog = ${bastille_jail_log};
  exec.start = '/bin/sh /etc/rc';
  exec.stop = '/bin/sh /etc/rc.shutdown';
  host.hostname = ${NAME};
  mount.devfs;
  mount.fstab = ${bastille_jail_fstab};
  path = ${bastille_jail_path};
  securelevel = 2;

  interface = ${bastille_jail_conf_interface};
  ${IPX_ADDR} = ${IP};
  ip6 = ${IP6_MODE};
}
EOF
}
```
And this is the code who handle some checks, last statement always obey and set ${INTERFACE} as expected:
```
if [ -z "${bastille_network_loopback}" ] && [ -n "${bastille_network_shared}" ]; then
    local bastille_jail_conf_interface=${bastille_network_shared}
fi
if [ -n "${bastille_network_loopback}" ] && [ -z "${bastille_network_shared}" ]; then
    local bastille_jail_conf_interface=${bastille_network_loopback}
fi
if [ -n "${INTERFACE}" ]; then
    local bastille_jail_conf_interface=${INTERFACE}
fi
```

Regards

usr/local/share/bastille/create.sh
JRGTH: Hi cedwards, this is just the default and how it works on Thickjail, if the user select ports initially when `bootstrapped` they should be presented in the jail seamlessly, however since the Thinjail nature and they have a shared base across the jails, this a basic requirement to map the wanted ports by the user on the jails by default, also this is only created if the `release/usr/ports` exist, just the same way the shared base is applied but with `read-write` permissions as expected.

IMO, this should be an one time handled `bootstrap` jail-built procedure and should be handled internally rather tan tempting basic requirements, between always up for suggestions.

Regards

PR #319, opened 2021-01-19
Update README / pf.conf

hackacad:
> pass out modulate state
modulate will only work on TCP connections.

Documentation changed to 
> pass out keep state


cedwards:
Bastille Blog and website documentation updated to match.

PR #326, opened 2021-01-27
zfssample

pvuchetich:
Add bastille.conf.zfssample for new install, update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,8 +233,7 @@ not using ZFS and can safely ignore these settings.
 bastille bootstrap
 ------------------
 Before you can begin creating containers, Bastille needs to "bootstrap" a
-release.  Current supported releases are 11.3-RELEASE, 12.0-RELEASE and
-12.1-RELEASE.
+release.  Current supported releases are 11.3-RELEASE, 12.0-RELEASE, 12.1-RELEASE, and 12.2-RELEASE
 
 **Important: If you need ZFS support see the above section BEFORE
 bootstrapping.**
@@ -250,6 +249,11 @@ ishmael ~ # bastille bootstrap 11.3-RELEASE
 **FreeBSD 12.1-RELEASE**
 ```shell
 ishmael ~ # bastille bootstrap 12.1-RELEASE
+```
+
+**FreeBSD 12.2-RELEASE**
+```shell
+ishmael ~ # bastille bootstrap 12.2-RELEASE
 ```
 
 **HardenedBSD 11-STABLE-BUILD-XX**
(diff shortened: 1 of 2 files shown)


pvuchetich:
This is my first pull request - I would appreciate being corrected / informed if I need to do something different.
The reasoning for this .zfssample file is that the original sample was confusing.
The original bastille.conf.sample included the annotation ## default: "${bastille_zfs_zpool}/bastille", but the default is actually "bastille".
When I used the exact string "${bastille_zfs_zpool}/bastille", it resulted in zfs filesystem creation errors.
```
original bastille.conf.sample
bastille_zfs_prefix="bastille"                                        ## default: "${bastille_zfs_zpool}/bastille"
```
The bastille.conf.zfssample uses "zroot" because that is the default zfs poolname on a FreeBSD 12.2-RELEASE installation.

The intention is that this could be copied to bastille.conf on a fresh default install of the most recent FreeBSD and just work.

hackacad (approved):
(1) yes,  bastille_zfs_mountpoint was already removed from Bastille.conf (it's not use anywhere in the code)
(2) yes, prefix should be `bastille` not `{bastille_prefix}/bastille`

hackacad:
You should change bastille.conf.sample (`bastille_zfs_prefix` defaults to...)  as well

hackacad:
I wouldn't create a .sample file for every use case.
It's just two commands. 
```
sysrc -f /usr/local/etc/bastille/bastille.conf bastille_zfs_enable=YES
sysrc -f /usr/local/etc/bastille/bastille.conf bastille_zfs_zpool=zroot
```

Maybe this could/should be added to docs.

hackacad (changes requested):
Feedback: Please create a separate PR for every issue. It's three changes in one PR.

pvuchetich:
ok - I interpreted the change as a single conceptual revision, but can work to break it up as follows, and submit multiple replacement PRs after resyncing to make sure there are no conflicts.
1. remove the .zfssample file (will not use separate file for sample per feedback)
2. correct the original .sample file (error in annotation text, not in actual variable)
3. Separate PR for README (to remove the reference to a variable no longer in use)
4. Add content to documentation with 2 line sysrc example, possibly additional descriptive text as needed.

Process question: Is it best to "Close/Reject" this PR, and submit all new ones, or to trim this one down to one change (e.g. just the README file), and submit multiple new ones?

cedwards:
In the end, whichever is easiest for you.
There are going to be new PR either way. 😄

hackacad:
There's way to modify the current PR, but the easiest way is to simply close it and create a new PR.

pvuchetich:
ok. Thanks. closing and will submit new, smaller ones.

PR #327, opened 2021-01-27
Minor md syntax fixes, adding missing spaces

fbarthez:

PR #328, opened 2021-01-27
fix formatting in network documentation

cedwards:
improve formatting for firewall documentation
update / standardize firewall docs


cedwards:
This should fix what was attempted in #327 .

PR #336, opened 2021-02-01
Fixing RDR for VNET jails

0x4D616E75:
Dynamic RDR isn't working because, `Bastille rdr` use `jls` to get IP address of jails. This isn't working when using VNET.

If VNET is used, ip address is extracted from inside the jail now.

### DIFF
--- a/usr/local/share/bastille/rdr.sh
+++ b/usr/local/share/bastille/rdr.sh
@@ -60,12 +60,14 @@ if [ -z "${JAIL_NAME}" ]; then
     error_exit "Jail not found: ${TARGET}"
 fi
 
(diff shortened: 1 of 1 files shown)


cedwards:
The rdr system is designed primarily for loopback networking where the container cannot be accessed directly.

VNET containers can handle their own firewalls and can be accessed by the outside network.

Please see:
https://bastille.readthedocs.io/en/latest/chapters/networking.html#virtual-network-vnet
https://bastillebsd.org/blog/2021/01/13/bastille-port-redirection-and-persistence/

bmac2:
per Christer in comments we will not do.  Closing.

Issue #330, opened 2021-01-28
[ENHANCEMENT] "Quiet" option for bastille cp


zilti:
**Is your feature request related to a problem? Please describe.**
I copy a quite large folder into a jail for testing, which creates a few thousand lines of output. And on top of that, it breaks the GitLab job display since there are too many lines.

**Describe the solution you'd like**
Something like a `-q` switch for `bastille cp`.

**Describe alternatives you've considered**
A workaround would be to directly copy it into the `/usr/local/bastille/jails/<jailname>/root` directory I guess, but that feels like an ugly hack.

**Additional context**
None

PR #350, opened 2021-02-18
Added option to have CP be quiet

zilti:
Adding a `-q` option for CP, resolving #330 .


cedwards:
I like this patch (thank you).
My only feedback is to standardize with `create` and others in putting the option after the sub-command vs at the end.
ie; `bastille cp -q TARGET source dest`

zilti:
Okay, this should be better.

JRGTH (approved):
This looks option `-q|--quite` looks ok being after `cp` command now. ;)

PR #354, opened 2021-02-26
Statistics2

yerrysherry:
This is a beter version then https://github.com/BastilleBSD/bastille/pull/310.

"bastille stat" gives some resource indication about jails. The figures are not 100% correct but now you have an idea about jails resources consumption.

JID  IP Address    Hostname or Jailname  Path                                                       Cpu (%)  Mem (%)  Rss (MB)  Vsz (MB)  Zfs disk
0    -                      test                                     -                                                            373.2    0               170             464
7    10.10.10.100  bastille1                             /usr/local/bastille/jails/bastille1/root    0           0                5                22            9.39M
8    10.10.10.50   ioc-dc1-minio-1                /jails/iocage/jails/dc1-minio-1/root      28.3      0.1              47              198          5.19G

### DIFF
--- a/usr/local/bin/bastille
+++ b/usr/local/bin/bastille
@@ -103,7 +103,7 @@ Available Commands:
   rename      Rename a container.
   restart     Restart a running container.
   service     Manage services within targeted container(s).
-  stat        Print jails statistics.
+  stat        List jails statistics.
   start       Start a stopped container.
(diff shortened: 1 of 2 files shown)


yerrysherry:
The first line gives information of the physical system. Each core is 100%, in this example, the total cpu is 400% = 4 cores

PR #355, opened 2021-02-26
Add preliminary support for MidnightBSD.

laffer1:


usr/local/share/bastille/pkg.sh
```diff
@@ -47,6 +47,10 @@ fi
 
 for _jail in ${JAILS}; do
     info "[${_jail}]:"
-    jexec -l "${_jail}" /usr/sbin/pkg "$@"
+    if [ -f ${_jail}/usr/sbin/pkg ]; then
```
cedwards, line 50: This looks odd to me. `${_jail}` should just be name so `${_jail}/usr/sbin/pkg` isn't a full path to evaluate?

laffer1: You're right, I pushed a fix up.

JRGTH (approved):
This small changes are working as expected on my quick test.

Regards!

PR #364, opened 2021-03-23
Bastille bootstrap list

hackacad:
new feature:
`bastille bootstrap list`

(and add RC3/RC4 as valid release, no extra PR for that)


cedwards:
`bastille list releases` already does this.

PR #377, opened 2021-04-23
Fixed #368 allowing mixed case YES values for bastille_zfs_enable

### LINKED ISSUE #368: bastille_zfs_enable="yes" , lowercase value should be parsed
in rc.conf lowercase value are also parsed.

indgy:
It took me quite a while to realise the `bastille_zfs_enable` value had to be UPPERCASE. This adds a function to capitalise any case 'yes' value, it also avoids touching the logical if statements in the various files. I've added the necessary code in any file where `bastille_zfs_enable` is used. It may be preferred to source a normalise.sh file at the top of every script, just after sourcing bastille.conf to avoid this cruft and ensure it's called in any new scripts, if the normalise method is preferred I'll submit another fix.

### DIFF
--- a/usr/local/share/bastille/bootstrap.sh
+++ b/usr/local/share/bastille/bootstrap.sh
@@ -35,6 +35,10 @@ usage() {
     error_exit "Usage: bastille bootstrap [release|template] [update|arch]"
 }
 
+# Capitalise bastille.conf $bastille_zfs_enable value
+capitalise_conf_yes $bastille_zfs_enable
+bastille_zfs_enable=$answer
+
 # Handle special-case commands first.
 case "$1" in
 help|-h|--help)
--- a/usr/local/share/bastille/clone.sh
+++ b/usr/local/share/bastille/clone.sh
@@ -35,6 +35,10 @@ usage() {
     error_exit "Usage: bastille clone [TARGET] [NEW_NAME] [IPADRESS]"
 }
 
+# Capitalise bastille.conf $bastille_zfs_enable value
+capitalise_conf_yes $bastille_zfs_enable
+bastille_zfs_enable=$answer
+
 # Handle special-case commands first
 case "$1" in
 help|-h|--help)
--- a/usr/local/share/bastille/common.sh
+++ b/usr/local/share/bastille/common.sh
@@ -48,3 +48,13 @@ info() {
 warn() {
     echo -e "${COLOR_YELLOW}$*${COLOR_RESET}"
 }
+
+# Used before comparing bastille.conf arguments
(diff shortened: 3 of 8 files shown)


chriswells0 (commented):
I like the general idea here, but it makes me wonder why we don't go ahead and convert all the values to uppercase once at the very beginning (/usr/local/bin/bastille). In addition, it would be great to error out if any required config values are empty or not set.

chriswells0:
After leaving the last comment, I looked at what it would take to do this in one place instead of many. The problem was that we were running the command scripts with `exec`, so they all had to include the common script and conf file. I created PR #397 to address that.

JRGTH:
Hello, another suggested simple solution is to use `[Yy][Ee][Ss]` instead:

```
$ case yes in [Yy][Ee][Ss]) echo 'WORKS';; esac
WORKS
$ case YES in [Yy][Ee][Ss]) echo 'WORKS';; esac
WORKS
$ case yEs in [Yy][Ee][Ss]) echo 'WORKS';; esac
WORKS
$ case YeS in [Yy][Ee][Ss]) echo 'WORKS';; esac
WORKS
```
Regards

chriswells0:
Yes, definitely. I just prefer knowing the data is always formatted than to expect everyone to test it correctly every time.

X86BSD:
I'd just look to see what FreeBSD does for rc.conf. sysrc can use both "YES" or "yes" when setting knobs in rc.conf. I wouldn't reinvent the wheel, just do what the project does.

bdrewery:
> I'd just look to see what FreeBSD does for rc.conf. sysrc can use both "YES" or "yes" when setting knobs in rc.conf. I wouldn't reinvent the wheel, just do what the project does.

From /etc/rc.subr
```
#
# checkyesno var
#       Test $1 variable, and warn if not set to YES or NO.
#       Return 0 if it's "yes" (et al), nonzero otherwise.
#
checkyesno()
{
        eval _value=\$${1}
        debug "checkyesno: $1 is set to $_value."
        case $_value in

                #       "yes", "true", "on", or "1"
        [Yy][Ee][Ss]|[Tt][Rr][Uu][Ee]|[Oo][Nn]|1)
                return 0
                ;;

                #       "no", "false", "off", or "0"
        [Nn][Oo]|[Ff][Aa][Ll][Ss][Ee]|[Oo][Ff][Ff]|0)
                return 1
                ;;
        *)
                warn "\$${1} is not set properly - see rc.conf(5)."
                return 1
                ;;
        esac
}
```

bmac2:
closing .  Thank you for the PR.

PR #380, opened 2021-05-15
fixed displaying issue

noracenofun:
fixed wrong display if there are no jails existing

```
find: /usr/local/bastille/jails/*/root/etc: No such file or directory
 JID  State  IP Address  Published Ports  Hostname  Release  Path
```
find issue fixed
```
 JID  State  IP Address  Published Ports  Hostname  Release  Path
```

directory `/usr/local/bastille/jails` not exist
```
unfortunately there are no jails here (/usr/local/bastille/jails)
```


cedwards:
I've also discovered that `os-release` is only available on 13.0-RELEASE and any jails older than that fail on that query.

noracenofun:
Yes or pkg/port sysutils/etc_os-release has been installed. I did. 😉

PR #383, opened 2021-05-25
Add support for codeberg.org

jan0sch:
This adds support for installing templates from codeberg.org which is a
free git hosting platform using gitea.


hackacad:
@cedwards agreed to add support for any git server. 
duplicate of #381

jan0sch:
Nice! :heart:  So I guess this one can be closed, right?

PR #384, opened 2021-05-25
Don't include jail path twice

gtz42:
In ed50e3fa041b35f274a30e0f2ac425252485e7f0 _jailpath was updated to include the full path on the host.
The test and if required mkdir are now done with the jail’s root directory perpended to that full path.


chriswells0 (commented):
It looks like this bug bit 2 people at the same time: #395 also resolves this, but it adds a fix for the fact that calling `mount` on all jails has the same "double path include" issue inside the loop.

cedwards:
Thank you for fixing this.

PR #386, opened 2021-05-27
Fixes #385 Ipv6 rdr support

### LINKED ISSUE #385: [ENHANCEMENT] ipv6 rdr support
Right now when creating a jail with both a local ipv4 and ipv6 address, only the ipv4 address will be considered for rdr rules by `bastille rdr`.

When a jail has both an ipv4 and ipv6 address, each should get rdr rules. I have not tested an ipv6 only jail but after reading https://github.com/BastilleBSD/bastille/blob/master/usr/local/share/bastille/rdr.sh I know it would not work either.

A static pf rule like `rdr  pass  on  $ext_if  inet6  proto  tcp  from  <internet>  to  <myself>  port  { http, https } ->  ::ffff:10:0:0:2` works around the problem but is cumbersome to maintain.

**Additional context**
I am fluent in shell and more than willing to write a patch then submit a pull request to add this feature. If you are interested, just let me know. Thanks for writing and maintaining such an amazing piece of software!

Have a great day.

adyxax:
This fixes the referenced issue, and a bunch of other stuff I encountered along the way : 
- When trying to add an rdr rule with a vnet jail no explicit error message was issued, I added one.
- I added support for ipv6 to `bastille rdr`.
- I fixed stop.sh to avoid calling bastille rdr directly which now raised an error message with vnet jails
- I fixed the jails pf table manipulations which did not support ipv6 addresses, nor vnet jails.
- I updated the documentation on rdr to make it explicit this mechanism is only intended for non-vnet jails. I wanted to add a similar note for the jails pf table but could not find existing documentation, if you have pointer on how you want to do that do not hesitate to ask.

I grouped the fixes in this one pull request but I can split the work if you prefer, do not hesitate to ask for modifications/enhancements.


adyxax:
I updated this PR following the merge of #394 and tested the changes again without running into any issue.

bmac2:
this PR is very stale and off of the old code base from 3 years ago.  Please if this is still a valid request, please update this to the current codebase and resubmit for review.

bmac2:
assigned this to myself to code up in new codebase

bmac2:
will be wrapped into ipv4/6 support throughout the system which is on the roadmap.

PR #388, opened 2021-06-06
Add support for BSD grep in list command, no need extended regex.

fudanchii:
wildcard `*` only supported in GNU grep, and since we're not specifying that GNU grep as dependency, was thinking we should support standard BSD grep too.
Screenshot from 2021-06-06 20-16-40

I think we could use better display here with more information. Willing to work on it if there is some suggestion.

### DIFF
--- a/usr/local/share/bastille/list.sh
+++ b/usr/local/share/bastille/list.sh
@@ -158,7 +158,7 @@ if [ $# -gt 0 ]; then
         rctl -h jail:
         ;;
     import|imports|export|exports|backup|backups)
-        ls "${bastille_backupsdir}" | grep -Ev "*.sha256"
+        ls "${bastille_backupsdir}" | grep -v ".*\.sha256"
     exit 0
     ;;
     *)


chriswells0 (approved):
This change looks good to me.

I don't like the "*.sha256" approach and I'm surprised it worked at all, but I'm curious what BSD distro/version you're on since it did work for me when I tried it. I'm using FreeBSD 12.2 and it says the default is GNU grep:
```
$ which grep
/usr/bin/grep
$ grep --version
grep (GNU grep) 2.5.1-FreeBSD
```

fudanchii:
oh, interesting, is it because mine is a new install? checked and seems yes, [13.0 moved to BSD grep by default](https://www.phoronix.com/scan.php?page=news_item&px=FreeBSD-13.0-RELEASE)
```
$ uname -rs
FreeBSD 13.0-RELEASE-p1
$ grep --version
grep (BSD grep, GNU compatible) 2.6.0-FreeBSD
$ /usr/local/bin/grep --version
grep (GNU grep) 3.6
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and others; see
<https://git.sv.gnu.org/cgit/grep.git/tree/AUTHORS>.
```

chriswells0:
Yes, that's what happened. I was listening to [BSDNow episode 406](https://www.bsdnow.tv/406) last night (after my earlier comment) and it came up in the listener questions. :)

Now I wonder how much other code will be impacted by this.

cedwards:
Is this fix backwards compatible? Works on 13 and 12?

fudanchii:
Yes, I believe so, the fix using regex syntax that should work for both gnu and bsd grep. As shown in screenshot above.

cedwards:
Looks like this was fixed in #413 .

Thank you for pointing this out and offering a fix.

PR #390, opened 2021-06-09
Lint bootstrap.sh

aeifn:
Lint with https://github.com/koalaman/shellcheck/wiki

### DIFF
--- a/usr/local/share/bastille/bootstrap.sh
+++ b/usr/local/share/bastille/bootstrap.sh
@@ -50,9 +50,7 @@ if [ "$(sysrc -n zfs_enable)" = "YES" ] && [ ! "${bastille_zfs_enable}" = "YES" 
         no|No|n|N|"")
             error_exit "ERROR: Missing ZFS parameters. See bastille_zfs_enable."
             ;;
-        yes|Yes|y|Y)
-            # continue
-            ;;
+        yes|Yes|y|Y) ;;
     esac
 fi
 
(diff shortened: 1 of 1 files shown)


usr/local/share/bastille/bootstrap.sh
```diff
@@ -336,7 +336,7 @@ bootstrap_template() {
             $(which git) clone "${_url}" "${_template}" ||\
                 error_notify "Clone unsuccessful."
         elif [ -d "${_template}/.git" ]; then
-            cd "${_template}" && $(which git) pull ||\
+            (cd "${_template}" && $(which git) pull) ||\
```
chriswells0, line 339: More of a "previous code" question than about your change: why are we doing `$(which git)` here (and 4 times)? If that finds _git_, then it would be found by simply calling `git`. Seems like an unnecessary subshell. The first time it's used makes sense to me since `-x` needs to know the path to the file, but the second time could be an `else` and the other 2 times could use `-C`.

On the change itself, it also seems like an unnecessary subshell. Git can be ran in any directory with `-C`. I believe all we need is this:
```
git -C "${_template}" pull
```

usr/local/share/bastille/bootstrap.sh
```diff
@@ -253,21 +253,21 @@ bootstrap_release() {
...
                     fi
                     error_exit "Bootstrap failed."
                 fi
 
                 ## fetch for missing dist files
                 if [ ! -f "${bastille_cachedir}/${RELEASE}/${_archive}.txz" ]; then
-                    fetch "${UPSTREAM_URL}/${_archive}.txz" -o "${bastille_cachedir}/${RELEASE}/${_archive}.txz"
-                    if [ "$?" -ne 0 ]; then
+                    if ! fetch "${UPSTREAM_URL}/${_archive}.txz" -o "${bastille_cachedir}/${RELEASE}/${_archive}.txz";
+                    then
```
chriswells0, line 270: Not a big deal, but it would be nice to leave `then` at the end of the line for consistency.

chriswells0 (commented):
It's nice to see clean-up work come in.

usr/local/share/bastille/bootstrap.sh
aeifn: Actually, it's better to say: `if ! which -s git;` instead `if [ ! -x "$(which git)" ];`
(-s means silent)

cedwards:
Thank you!

PR #394, opened 2021-06-09
Fixes #338 Network access on jail startup 

### LINKED ISSUE #338: [ENHANCEMENT] Network access on jail startup
A jail with nginx takes a long time to start. With OCSP stapling enabled the startup blocks on name lookup.
Examining start.sh, the IP address of the jail is added to the <jails> table of PF only after startup causing the lookup to fail.

Move the addition to the <jails> table to immediately before the jail call.

I worked around this by adding a static NAT rule for this jail.

**Additional context**
Used bastille-0.8.20210115 from latest pkg with private networking.

adyxax:
This fixes the referenced issue.

This PR will apply cleanly on master but would conflict with my other pull request #386 where I added ipv6 support. Which ever you decide to merge first I will rework the other one so that it applies cleanly.

I updated the PR to remove duplicate code, a jail ip was already fetched in the code right above. I am not a fan of the grep|awk|sed in place and I left this code as is, but I can improve on it if you prefer.

### DIFF
--- a/usr/local/share/bastille/start.sh
+++ b/usr/local/share/bastille/start.sh
@@ -83,6 +83,8 @@ for _jail in ${JAILS}; do
                 error_notify "Error: IP address (${ip}) already in use."
                 continue
             fi
+            ## add ip4.addr to firewall table:jails
+            pfctl -q -t jails -T add "${ip}"
         fi
 
         ## start the container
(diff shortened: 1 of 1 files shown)


chriswells0 (commented):
Before this change, the jail was only being added to the table if it was using the loopback adapter. Now it's always added. Do VNET/other jails need to be in this firewall table?

adyxax:
I think that by definition vnet jails will never have ip4.addr set, one of their perks being to manage the network independently from inside the jail.
The `bastille create` command reflects that when used to create a vnet jail, it will not configure an ip4.addr.

cedwards:
Thank you!

PR #397, opened 2021-06-12
Allow any case for bastille_zfs_enable value.

### LINKED ISSUE #368: bastille_zfs_enable="yes" , lowercase value should be parsed

chriswells0:
Allow any case for bastille_zfs_enable value.
Make common.sh and bastille.conf available to all command scripts.

Closes #368.


chriswells0:
Note that I did not do extensive testing of all the command scripts. I tested a small subset like listing jails and installing/auditing packages in a single jail.

PR #409, opened 2021-06-29
Update template.rst

tobybryans:
* Made the difference between old style and post 0.7.x style configuration more clear
* Fixed the template link at the top of the page
* Added more explanation about file layout for the Bastillefile and overlay structure
* Added some notes about the RDR hook
* Changed the command example to add usr as an OVERLAY to use the new style configuration

### DIFF
--- a/docs/chapters/template.rst
+++ b/docs/chapters/template.rst
@@ -1,53 +1,84 @@
 ========
 Template
 ========
-Looking for ready made CI/CD validated [Bastille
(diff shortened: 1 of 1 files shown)


cedwards:
Thank you for this. My only feedback is we're standardizing on Bastille sub-commands for templates which means we'll want to replace `OVERLAY` with `CP`.

cedwards:
I ended up improving the template documentation with the last release. Thank you for putting this together. It was helpful in pointing out what needs clarification; closing.

PR #413, opened 2021-07-09
fix list.sh for BSD grep

hackacad:
This PR fixes list.sh import.

>grep: repetition-operator operand invalid


JRGTH:
Thanks for fixing, the PR f4738cb works as intended indeed.

PR #415, opened 2021-07-12
Revision for create cmd and cleanup, fixed regressions and enhanced usage

JRGTH:
This update fixes for the previous regressions on the `create.sh`command and code cleanup.
This also enhances the usage display for user convenience.


cedwards:
Thank you.

PR #440, opened 2021-08-17
Resolve Hostname to Localhost

AverageHumanoid:
Well this seems to do want I want in #439, everytime the container starts it will update the `/etc/hosts` file to resolve the hostname to localhost. If someone can come up with a more elegant solution then please use that instead.

### DIFF
--- a/usr/local/share/bastille/start.sh
+++ b/usr/local/share/bastille/start.sh
@@ -104,6 +104,9 @@ for _jail in ${JAILS}; do
                 bastille rdr "${_jail}" ${_rules}
             done < "${bastille_jailsdir}/${_jail}/rdr.conf"
         fi
+        
+        ## update hosts file
+        sed -i '' '/localhost/ s/[^[:blank:]]\{1,\}$/'${_jail}'/' "${bastille_jailsdir}/${_jail}/root/etc/hosts"
     fi
     echo
 done


AverageHumanoid:
This will insert an entry like `10.17.89.50 alcatraz` below `127.0.0.1 localhost` and keep it updated when the jail is started and work for VNET jails.

hackacad:
I don’t think that this should be done by default, especially not on every reboot. 

It could be accomplished by a template, which can be easily replaced but doesn’t require tracking the upstream code. (Some deployments don’t want or might even reject unexpected hostname resolutions)
Name resolutions should be done by a DNS server anyway. (This PR would override DNS...)

There are cases where you don’t want that exact name resolution, so the jail manager shouldn’t force any settings it expects.

AverageHumanoid:
Other container managers automatically do this so when the apps I tried to run using Bastille wouldn't start because it couldn't resolve the hostname, I was quite confused. My original solution was to map hostname to localhost but some people in the Bastille discord server had problems with that so I changed it to jail IP address like they suggested. If you prefer to not have it running by default then how about only running this if a `-u` flag is included in `bastille start`?

AverageHumanoid:
New approach will only update if you run 
```
bastille hosts alcatraz
bastille start alcatraz hosts
bastille restart alcatraz hosts
```
and in Bastillefile.

AverageHumanoid:
I kind of copied the functionality of the `update`command since you can run `bastille update 12.2-RELEASE` as well as `bastille bootstrap 12.2-RELEASE update`. Should I revert `start` and `restart` so only the `hosts` command work for this?  Should I rename the `hosts` command to something else?

AverageHumanoid:
Now this will only update a jail's `/etc/hosts` file if run
```
bastille hosts alcatraz
bastille hosts ALL
```
and in Bastillefile.

PR #444, opened 2021-09-11
PKG.SH::CHANGED:: use host's pkg to install packages in jails

frikilax:
This change has several benefits:
- avoid installing the pkg system in jails
- allow potential ENVVARs to be taken into account when executing the bastille command

### DIFF
--- a/usr/local/share/bastille/pkg.sh
+++ b/usr/local/share/bastille/pkg.sh
@@ -31,7 +31,7 @@
 . /usr/local/share/bastille/common.sh
 
 usage() {
-    error_exit "Usage: bastille pkg TARGET command [args]"
+    error_exit "Usage: bastille pkg TARGET [-P|--pkg] command [args]"
 }
 
 # Handle special-case commands first.
@@ -39,6 +39,17 @@ case "$1" in
 help|-h|--help)
     usage
     ;;
+esac
+
+## initialize options
+USE_HOST_PKG=""
+
+## handle additional options
+case "${1}" in
+    -P|--pkg)
+        USE_HOST_PKG="1"
+        shift
+        ;;
 esac
 
 if [ $# -lt 1 ]; then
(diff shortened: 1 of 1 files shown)


hackacad (commented):
What if the hosts OS release is different from the jail?
This will not work on heterogeneous Hosts (host version != jail version)

JRGTH:
> What if the hosts OS release is different from the jail?
> This will not work on heterogeneous Hosts (host version != jail version)

I was about to chime in as well.

We have to take into consideration Embedded NAS'ses who are too stripped down as an example, this will prevent those host to update packages through `bastille`, maybe an user variable can be an alternative, to override such behavior if set.

frikilax:
> 
> 
> What if the hosts OS release is different from the jail?
> This will not work on heterogeneous Hosts (host version != jail version)

I tried using the method on a HardenedBSD 12 host, updating FreeBSD 12 and 13 releases and the result is pkg using the correct configuration file(s) and updating the right (jail's) database.



> We have to consider Embedded systems who may have the OS stripped down and some don't have pkg tools etc., this will prevent those systems from update packages through `bastille`, maybe an user variable can be an alternative, to override such pkg behavior when set.

You got a point, i was thinking the other way around (creating "limited" jails without pkg system), but it's also not guaranteed pkg is installed on the host system.

telser:
Could this instead be added as another option? I would certainly find it useful and prefer to not install pkg in most of my jails. Having the system aware tooling with jails, like pkg, is one of my favorite things about them.

frikilax:
> Could this instead be added as another option? I would certainly find it useful and prefer to not install pkg in most of my jails. Having the system aware tooling with jails, like pkg, is one of my favorite things about them.

yes, this seems like the best way to integrate the modification

cedwards:
I like supporting this as an option. My only feedback is to shift the option back to the sub-command to match options used in other commands. (ie; bastille create -V ..., bastille pkg -P ...)

Do you mind updating this to expect the argument before the target?

frikilax:
obsoleted by #474

PR #459, opened 2021-11-15
Add a description to the host vnet interface

robarnold:
This makes it much easier to understand `ifconfig` when run on the jail host. Not sure how to test this, but manually appyling this via `bastille edit` to an existing vnet jail had the right effects when I restarted the jail.


cedwards:
I like this. Thank you.

Let me do some testing in my lab before merging.

PR #474, opened 2021-12-23
jls path fix + support using host package manager

cedwards:
This patch completes a previous attempt at defining full path to jls binary.

Also included is new support for host package manager using the `pkg -P` or `pkg --pkg` options.


cedwards:
This completes the work proposed in #444. Thank you to @frikilax for the suggestion.

PR #480, opened 2022-01-10
Import iocage basejails as Bastille thin jails

robarnold:
I think these are the same concept, but with slightly different execution.
The main idea is to have a central base system that is shared (readonly) among
multiple jails. When this base system is updated, all the jails that reference
it immediately see the new system files.

This is implemented in iocage as a set of individual zfs mounts, one per
system directory. In Bastille, each system directory is symlinked into a
subdirectory of a single zfs mount.

My test plan here was to import an exported iocage basejail and verify that
its Bastille version has the right fstab and symlink changes:
```
Validating file: ssl_2021-11-19.zip...
File validation successful!
Importing 'ssl' from foreign compressed .zip archive.
Archive:  ssl_2021-11-19.zip
 extracting: ssl_2021-11-19
 extracting: ssl_2021-11-19_root
Receiving ZFS data stream...
Generating jail.conf...
Updating symlinks...
ln: usr/src: Directory not empty
Warning: directory usr/src on imported jail was not empty and will not be updated by Bastille
Container 'ssl' imported successfully.
```

### DIFF
--- a/usr/local/share/bastille/import.sh
+++ b/usr/local/share/bastille/import.sh
@@ -181,15 +181,21 @@ generate_config() {
         if [ -n "${JSON_CONFIG}" ]; then
             IPV4_CONFIG=$(grep -wo '\"ip4_addr\": \".*\"' "${JSON_CONFIG}" | tr -d '" ' | sed 's/ip4_addr://')
             IPV6_CONFIG=$(grep -wo '\"ip6_addr\": \".*\"' "${JSON_CONFIG}" | tr -d '" ' | sed 's/ip6_addr://')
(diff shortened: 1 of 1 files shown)


JRGTH:
I've tested this PR and is working as expected, thank you!

Import output:
```
# bastille import ioc_basejail_2022-01-13.zip
Validating file: ioc_basejail_2022-01-13.zip...
File validation successful!
Importing 'ioc_basejail' from foreign compressed .zip archive.
Archive:  ioc_basejail_2022-01-13.zip
 extracting: ioc_basejail_2022-01-13  
 extracting: ioc_basejail_2022-01-13_root  
Receiving ZFS data stream...
Generating jail.conf...
Updating symlinks...
ln: usr/src: Directory not empty
Warning: directory usr/src on imported jail was not empty and will not be updated by Bastille
Container 'ioc_basejail' imported successfully.
```

Just a tiny update on the `jail.conf` as expected and good to go!

PR #531, opened 2022-07-18
bugfix for cmd when using ALL target

cedwards:
Simple fix for addition of Linux support when using ALL target.


JRGTH (approved):
Hey cedwards good catch, since I was testing Linux jails once at a time didn't noticed this bug. now the command works over a batch of Linux jails as well as for single/targeted jails.

Fix working as expected.

bmac2 (approved):

bmac2 (approved):
approved

PR #534, opened 2022-07-24
Make parsing on jail.conf more robust

cqexbesd:
jail.conf has a few tricky things about it's format (such as supporting
variable expansion) so it is easiest to get jail to do the parsing for us.

The changes to `get` required changes to `set` as well as we are no longer calling grep so that became a single awk script as well.

I hope there are some good tests for this code! One thing that has changed is the stripping of quotes on output, if they were present. I think this is better than sometimes having (and makes things easier for me for an upcoming change). I didn't see anywhere that broke existing code but I'm just a beginner.

### DIFF
--- a/usr/local/share/bastille/config.sh
+++ b/usr/local/share/bastille/config.sh
@@ -35,6 +35,15 @@ usage() {
     error_exit "Usage: bastille config TARGET get|set propertyName [newValue]"
 }
 
+# we need jail(8) to parse the config file so it can expand variables etc
+print_jail_conf() {
+
+    # we need to pass a literal \n to jail to get each parameter on its own
+    # line
+    jail -f "$1" -e '
+'
+} 
+
 # Handle special-case commands first.
 case "$1" in
 help|-h|--help)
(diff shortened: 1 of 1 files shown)


cqexbesd:
I have spotted a bug so this isn't ready to be merged yet. Will push an update soon.

cqexbesd:
OK I think its good to go now. I missed adding a new line to the config if the property wasn't already set.

cqexbesd:
One other change in behaviour is that "not set" is now just printed and doesn't go via `warn`. Is this a problem? Let me know if you want that fixed.

cqexbesd:
> One other change in behaviour is that "not set" is now just printed and doesn't go via `warn`. Is this a problem? Let me know if you want that fixed.

OK I realise I was just being lazy so I had fixed that now as well. Hopefully that's the last of it!

JRGTH:
Just basic tested with/without quotes and with/without spaces, and with pipe `|` character, and working here as expected.

P.S. Letting `jail` the parsing with variable expansion seems nice to have.

Sample output [HERE](https://dpaste.org/0gF4t)

JRGTH (commented):
Tested the basics and the `jail` parsing and var expansion seems to be working here as expected.
Between more testing is welcome.

Issue #369, opened 2021-03-24
[ENHANCEMENT]availability of  ip4/6 = inherit mode


s1dh:
an inherit mode for sharing host IP address would be usefull

It is provided by jail() and even iocage too

cedwards:
Can you provide examples and/or usage?

s1dh:
inherit is a possible value for  ip4/ip6 core parameter of  jail(8) to use host's listening address, It is very handy to quickly test jailed services (upgrade/migration) on hosted testing servers, no need to setup IP for the jail so you can zfs send/recv jail from one test server to another very easily.

cedwards:
I threw together a patch to support ip4/ip6 inherit. Will do a bit more testing before I merge.

bmac2:
@cedwards was this patch merged previously?????  What PR was it under?

jazzi:
Here is my usage that needs this feature @cedwards 

I have a very simple home server, just two services: NFS and Mpd music server, to simplify life I don't want to remember two IPs, just want all share the same IP with the host. Lazy? but easy.

bmac2:
PR #830 merged

PR #535, opened 2022-07-24
Allow ip4/ip6 inherit

cqexbesd:
Allow users to set `ip4` or `ip6` to `inherit`.

This isn't perfect as:

1) there isn't support for creating these jails in the first place (I create one close to what I want and edit `jail.conf` by hand)
2) there is an assumption in Bastille that jail will have an IP, and that `pf` will be in use so things like `stop`, while working, also emit `Jail IP not found: nnnn` from `rdr.sh`

However it seems to work and might satisfy #369 and #405.

I tried to address the `stop` IP assumption but I wasn't sure enough about the circumstances when there will be `pf` rules in place. Perhaps someone who understands all he permutations could produce a table of the different allowed values of ip[46]* and what it means for other networking things (pf rules, vnet interfaces, pairs etc). Then there could be a common set of functions other scripts could call to check if they should be doing any of that stuff.

NB this PR assumes #534 gets pulled first - then this becomes just the one commit.

### DIFF
--- a/usr/local/share/bastille/start.sh
+++ b/usr/local/share/bastille/start.sh
@@ -35,6 +35,21 @@ usage() {
     error_exit "Usage: bastille start TARGET"
 }
 
+# indicate if an IP configurtaion value (e.g. a value given for ip4 or ip6)
+# requires extra configuration external to the jail
+#
+# success if it does, failure if it does not
+ip_require_config() {
+
+    case "${1}" in
+        disable|inherit|"not set")
+            return 1
+            ;;
+    esac
+
+    return 0
+}
+
 # Handle special-case commands first.
 case "$1" in
 help|-h|--help)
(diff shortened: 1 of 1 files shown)


arrowd:
I found it disappointing that it is not possible to just inherit ip4 and ip6 from the host. Looking forward for this PR get merged in.

bmac2:
534 has been previously merged.

bmac2:
@cqexbesd  I want to merge this but need you to update the code to resolve the conflicts.  Sorry it took us this long to get  to this.

cqexbesd:
Rebased

bmac2:
will review with @cedwards this week and get merged.

cedwards:
I like this but let's work on expanding it to full support before we merge anything. In other words, let's add support for creating jails using inherit and make it a first class citizen.

For my own reference, I would appreciate any real examples of using inherit (honestly not a feature I ever personally use, hence the lack of current support). Also, any docs or code on how iocage or others handle inherit-based jails would be nice for me to better understand.

bmac2:
@cqexbesd  can you give some use cases and examples of this so we understand the ask and can determine how to make this happen?  @cedwards asked for examples a year ago so if this is not required anymore letus know so we can close it.

@yaazkal

bmac2:
#830 has this PR in it and will be merged soon.

@yaazkal @tschettervictor

bmac2:
closing as PR #830 makes this one obsolete.

PR #543, opened 2022-09-02
Updated "export" sub-command docs (0.9.20220216)

tofazzz:
"Updated with syntax on 0.9.20220216 as if running just bastille export TARGET it errors with:

"Error: Stream can not be written to a terminal.
You must redirect standard output.

Error: An export option is required, see 'bastille export, otherwise the user must redirect to standard output."

### DIFF
--- a/docs/chapters/subcommands/export.rst
+++ b/docs/chapters/subcommands/export.rst
@@ -16,4 +16,14 @@ can be exported only when the jail is not running.
 
 .. code-block:: shell
 
-  Usage: bastille export TARGET
+  Usage:  bastille export | option(s) | TARGET | PATH
+
+Available options are:
+
+         --gz       -- Export a ZFS jail using GZIP(.gz) compressed image.
+    -r | --raw      -- Export a ZFS jail to an uncompressed RAW image.
+    -s | --safe     -- Safely stop and start a ZFS jail before the exporting process.
+         --tgz      -- Export a jail using simple .tgz compressed archive instead.
+         --txz      -- Export a jail using simple .txz compressed archive instead.
+    -v | --verbose  -- Be more verbose during the ZFS send operation.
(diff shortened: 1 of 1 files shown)


cedwards:
Thank you

JRGTH:
Thanks for this doc update.

As expected an `export` option is required for a built-in/predefined export archive in which can be later re-imported, commonly used for basic local backups but not limited to.

While no `export` option means the user has to redirect the custom export regarding the admin needs, from a local backup with an alternate compression method, to a remote computer jail replication(advanced).

Additionally, setting a `${bastille_export_options}` in the bastille configuration file or setting an environment variable, will set a predefined export option with the user defined options, this will override the default redirection behavior, in my case as example I prefer to export with the below configuration in my home lab since is faster than `.txz`, safe and the size is just a tad bigger but not by much :
```
bastille_export_options="--safe --gz"
```

From the `export.sh` command comment:
```
if [ -n "${bastille_export_options}" ]; then
    # Overrides the case options by the user defined option(s) automatically.
    # Add bastille_export_options="--optionA --optionB" to bastille.conf, or simply `export bastille_export_options="--optionA --optionB"` environment variable.
    # To restore the standard case options, empty bastille_export_options="" in bastille.conf, or `unset bastille_export_options` environment variable.
    # Reference "/bastille/issues/443"
```

As denoted if the `${bastille_export_options}` is set the standard case options will be overridden and an info message will be displayed about the user export options choice like in this example:
```
root@mserver: ~# bastille export syncthing
Default export option(s): '--safe --gz'
Exporting 'syncthing' to a compressed .gz image...
Creating temporary ZFS snapshot for export...
Sending ZFS data stream...
 70.0%
Exported '/mnt/storage/bastille/backups/syncthing_2022-09-02-205636.gz' successfully.
``` 

Regards

Issue #597, opened 2023-07-19
[BUG] Start of VNET Jails only possible once 'bastille create -V' has been invoked


joh-ku:
**[MANDATORY] Describe the bug [MANDATORY]**
I want to migrate some Bastille Jails from Host A to Host B.
On host A I've exported the Jails with `bastille export --gz <JAIL>`.
On host B I've installed Bastille from ports and bootstrapped `13.2-RELEASE`. 
Then I've imported the first Jail.
The import succeeded, but when starting the Jail, I get the following error message:

```
jail: myjail: exec jib: No such file or directory
jail: myjail: jib addm bastille0 vmx0: failed
```
The problem is, that the [jib](https://cgit.freebsd.org/src/tree/share/examples/jails/jib?h=stable/13) script is only copied over to `/usr/local/bin/jib` when "[bastille create](https://github.com/BastilleBSD/bastille/blob/61e7084b335fa6fcdfa70988da57d5c365f431bc/usr/local/share/bastille/create.sh#L469-L477)" is invoked on a VNET enabled jail.
As this was not the case in my scenario, `jib` was missing after the Jail has been imported.

**[MANDATORY] Bastille and FreeBSD version (paste ``bastille -v && freebsd-version -kru`` output)**
0.9.20220714
13.2-RELEASE
13.2-RELEASE
13.2-RELEASE

**[MANDATORY] How did you install bastille? (port/pkg/git)**
Installed from ports.

**[optional] Expected behavior**
To avoid this, the availability of `jib` could be checked in [start.sh](https://github.com/BastilleBSD/bastille/blob/61e7084b335fa6fcdfa70988da57d5c365f431bc/usr/local/share/bastille/start.sh) instead.

bmac2:
fixed by @JRGTH in and merged to main.

PR #598, opened 2023-07-19
Validate jib during jail import

JRGTH:
This import enhancement will locate/install the jib script required by VNET jails.

Issue/Bug reference: https://github.com/BastilleBSD/bastille/issues/597


joh-ku:
Hi @JRGTH, thanks for the quick PR. For the avoidance of redundant code, wouldn't it be sufficient to check for `jib` in `start.sh` only?

JRGTH:
> 

Hi @joh-ku, The `jib` command should be installed just once upon a VNET jail creation or a VNET jail import(bug fix issued) as you've reported, unless manually installed by user.

Also yes the `jib` validation can be done in the `start.sh` command as well but if the user have lots of VNET jails this may lead for several `grep` command executions which is unnecessary.

joh-ku:
Understood, thank you for the clarification.

PR #605, opened 2023-09-07
Add rdr pf conf check

JRGTH:
Prevent shutdown/rdr command stalls if pf config is empty/non set.


tobiastom:
Just for the record: this did not make it into the release last week and it broke Bastille for me. 
After applying this change manually Bastille stopped again.

PR #608, opened 2023-09-19
Permit jail upgrades to use BETA and RC branches

tub5ta:
Allow the uplift of jails to test BETA and RC branches before release.

Currently to can bootstrap these branches with bastille but you cannot upgrade during the release cycle.


cedwards:
I like this.
Before we merge, let's extend to support bootstrapping BETA and RC releases.

PR #611, opened 2023-10-11
prep & cleanup for 0.10.20231013 release

cedwards:
A bunch of cleanup and changes for the 0.10.20231013 release. This also includes removing a bunch of old stuff from the README and instead pointing to the documentation. Going forward the README should remain lightweight and point to standard docs.


bmac2 (approved):
reviewed

PR #615, opened 2023-10-21
CIDR documentation on create command

bmac2:
documented that ONLY CIDR notation currently works for creating a jail instead of using  a netmask statement.

### DIFF
--- a/docs/chapters/subcommands/create.rst
+++ b/docs/chapters/subcommands/create.rst
@@ -24,11 +24,11 @@ address to the new system.
 
 .. code-block:: shell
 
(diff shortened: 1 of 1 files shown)


docs/chapters/subcommands/create.rst
```diff
@@ -22,6 +22,15 @@ bootstrapped release and a private (rfc1918) IP address.
 This command will create a 11.3-RELEASE container assigning the 10.17.89.10 ip
 address to the new system.
 
+.. code-block:: shell
+
+   ishmael ~ # bastille create alcatraz 13.2-RELEASE 10.17.89.113/27
```
cedwards, line 27: maybe let's default to `/24`

PR #627, opened 2023-11-05
Enhancement: optional ext_if6 macro in pf.conf to enable rdr for non-VNET dual-stack jails  

waseigo:
1. Added an optional parameter `ext_if6` for an IPv6 interface in `pf.conf` interface next to the default `ext_if` (for IPv4) in `bastille.conf`.
2. Adjusted `rdr.sh` so that rdr rules are also added and persisted for the `ext_if6` interface.

This makes it possible to have a non-VNET jail that can deal with packets coming from a Wireguard interface (IPv4) and from an Yggdrasil (IPv6) interface.

For example, I am running caddy in a non-VNET dual-stack jail, and this way I can have it reverse-proxy to other jails.

Prior to this change, if I didn't want to deal with VNET, I would have to have a separate non-VNET IPv6-only jail running caddy to deal with requests coming from the Yggdrasil interface.

This doesn't affect the creation of an IPv4-only jail.

I have not changed anything in `create.sh`. After creating an IPv4-only jail, I edit its `jail.conf` and add `ip6.addr = fd80:...` (in this case, a link-local address) and delete the line `ip6 = new;`

For an IPv6-only jail, an improvement to this would be to make `create.sh` set `ip4 = disabled;` (for which other scripts might need to be adjusted).

The changes to the "standard" `pf.conf` look like this:

```
ext_if=wg0
ext_if6=ygg0

set block-policy return
scrub in on { $ext_if, $ext_if6 } all fragment reassemble
set skip on lo

table <jails> persist
nat on $ext_if from <jails> to any -> ($ext_if:0)
nat on $ext_if6 from <jails> to any -> ($ext_if6:0)
rdr-anchor "rdr/*"

block in all
pass out quick keep state
antispoof for $ext_if inet
antispoof for $ext_if6 inet6
pass in inet proto tcp from any to any port ssh flags S/SA modulate state
pass in inet6 proto tcp from any to any port ssh flags S/SA modulate state
```

### DIFF
--- a/usr/local/bin/bastille
+++ b/usr/local/bin/bastille
@@ -45,6 +45,7 @@ bastille_conf_check
 . /usr/local/etc/bastille/bastille.conf
 # Set default values for config properties added during the current major version:
 : "${bastille_network_pf_ext_if:=ext_if}"
+: "${bastille_network_pf_ext_if:=ext_if6}"
 : "${bastille_network_pf_table:=jails}"
 
 ## bastille_prefix should be 0750
@@ -62,7 +63,7 @@ bastille_perms_check() {
 bastille_perms_check
 
 ## version
-BASTILLE_VERSION="0.10.20231013"
+BASTILLE_VERSION=b7d741b5cd3b0c758f0983fd9546e88fba0354d7
 
 usage() {
     cat << EOF
--- a/usr/local/etc/bastille/bastille.conf.sample
+++ b/usr/local/etc/bastille/bastille.conf.sample
@@ -54,6 +54,7 @@ bastille_export_options=""                                            ## default
 ## Networking
 bastille_network_loopback="bastille0"                                 ## default: "bastille0"
 bastille_network_pf_ext_if="ext_if"                                   ## default: "ext_if"
+bastille_network_pf_ext_if6="ext_if6"                                 ## default: "ext_if6"
 bastille_network_pf_table="jails"                                     ## default: "jails"
 bastille_network_shared=""                                            ## default: ""
 bastille_network_gateway=""                                           ## default: ""
--- a/usr/local/share/bastille/rdr.sh
+++ b/usr/local/share/bastille/rdr.sh
@@ -53,6 +53,7 @@ JAIL_NAME=""
 JAIL_IP=""
 JAIL_IP6=""
 EXT_IF=""
+EXT_IF6=""
 shift
 
 check_jail_validity() {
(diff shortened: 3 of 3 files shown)


usr/local/bin/bastille
```diff
@@ -45,6 +45,7 @@ bastille_conf_check
 . /usr/local/etc/bastille/bastille.conf
 # Set default values for config properties added during the current major version:
 : "${bastille_network_pf_ext_if:=ext_if}"
+: "${bastille_network_pf_ext_if:=ext_if6}"
```
cedwards, line 48: I assume this variable is meant to be `${bastille_network_pf_ext_if6:=ext_if6}`?

usr/local/bin/bastille
```diff
@@ -62,7 +63,7 @@ bastille_perms_check() {
 bastille_perms_check
 
 ## version
-BASTILLE_VERSION="0.10.20231013"
+BASTILLE_VERSION=b7d741b5cd3b0c758f0983fd9546e88fba0354d7
```
cedwards, line 66: This doesn't need to be included in the patch.

usr/local/bin/bastille
waseigo: Yes, you are correct... I fixed it locally but forgot to push.

usr/local/bin/bastille
waseigo: OK, will keep it in mind for next time!

waseigo:
@cedwards I am not terribly proficient at this (to put it mildly), so haven't run it through the gauntlet, so to say, to see if other things break.

usr/local/bin/bastille
yaazkal: @waseigo can you push the change so we can review it again?

usr/local/bin/bastille
yaazkal: @waseigo please then remove the change from your PR

tschettervictor:
Update on this?

I've redone much of the code on rdr.sh to allow setting an interface as well as setting a source and destination IP.

Check out PR #765 and test it to see if it will work for you. 

With it you can specify an interface to use with rdr so no need to add additional rules to pf.conf

waseigo:
> @waseigo can you test #765 to see if it works for you?

Hello -- in the meantime I've torn down that server instance and moved to Proxmox VE for my hosting needs, so I don't have a FreeBSD installation to test this with. Nevertheless, thank you for your work! I hope I can return to FreeBSD and Bastille one day.

tschettervictor:
@bmac2 We could probably close this PR then, and focus on the above.

bmac2:
The function in this one is in #765

PR #629, opened 2023-11-11
rcorder(8)-ed startup script

dsh2dsh:
With

```sh
bastille_enable="YES"
bastille_rcorder="YES"
```

in `/etc/rc.conf`, the script will start all jails, except jails with `KEYWORD: nostart` in jail.conf.

Example of `jail.conf` with `KEYWORD: nostart`:

```sh
# KEYWORD: nostart
jailname {
...
}
```

`PROVIDE:` is optional. Actually all `rcorder(8)` labels are optional, but we can use it to build jail dependencies. For instance, if we have jail `db` and jails `alfa` and `zeta`, we can configure it so both jails require jail `db`:

`alfa/jail.conf`:
```sh
# REQUIRE: db
alfa {
...
}
```

`zeta/jail.conf`:
```sh
# REQUIRE: db
zeta {
...
}
```

`db/jail.conf`:
```sh
# PROVIDE: db
db {
...
}
```

With this configuration jail `db` will start first and stop last.

### DIFF
--- a/usr/local/etc/rc.d/bastille
+++ b/usr/local/etc/rc.d/bastille
@@ -8,10 +8,19 @@
 
 # Add the following to /etc/rc.conf[.local] to enable this service
 #
-# bastille_enable (bool):          Set to NO by default.
-#               Set it to YES to enable bastille.
-# bastille_list (string):        Set to "ALL" by default.
-#               Space separated list of jails to start.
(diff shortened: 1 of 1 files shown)


cedwards:
Thank you @dsh2dsh . Nice to see you again 😄 .

PR #643, opened 2023-11-27
Adjust devfs_ruleset for hierarchical jails compatibility

draga79:
Hierarchical jails inherit the parent jail's permissions and don't support setting devfs_ruleset to a non-zero value. This update adds a check to determine if the script is running inside a jail. If so, it sets devfs_ruleset to 0 to comply with this constraint.

This will fix the problem described here: https://it-notes.dragas.net/2023/11/27/migrating-from-vm-to-hierarchical-jails-freebsd/

### DIFF
--- a/usr/local/share/bastille/create.sh
+++ b/usr/local/share/bastille/create.sh
@@ -165,10 +165,15 @@ EOF
 }
 
 generate_jail_conf() {
+    if [ "$(sysctl -n security.jail.jailed)" -eq 1 ]; then
+        devfs_ruleset_value=0
+    else
+        devfs_ruleset_value=4
+    fi
     cat << EOF > "${bastille_jail_conf}"
 ${NAME} {
-  devfs_ruleset = 4;
   enforce_statfs = 2;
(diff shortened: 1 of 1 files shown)


yaazkal:
Thanks @draga79 for the PR. We had an issue (#698) related to devfs rules that is fixed now. Maybe that could be the root cause. We would appreciate it if you could test again to check if this PR is still needed.

yaazkal:
Thanks @draga79 your PR looks like a great improvement for defaulting that configuration. Merging now.

Pashugan:
> Hierarchical jails ... don't support setting devfs_ruleset to a non-zero value

For searchability, this doesn't quite appear to be true. Setting the ruleset to the same value as the parent's jail does work as well as setting it to zero. Interestingly, the ruleset doesn't even need to be defined in the child jail.

That being said, I am very confused with this behaviour, and wonder if it is actually by design, and if there are plans to change it. Restricting the child jails further with a subset of devices would be more secure and logical.

Issue #647, opened 2023-12-22
[BUG] PKG Install method fails during setup, version 0.10.20231125 @ 14.0-RELEASE-p3 


em-winterschon:
**[MANDATORY] Describe the bug [MANDATORY]**
Setup stage `configure_zfs` fails when following the installation process from the docs, when using `pkg` method.

**[MANDATORY] Bastille and FreeBSD version (paste ``bastille -v && freebsd-version -kru`` output)**

```
bastille -v && freebsd-version -kru
0.10.20231125
14.0-RELEASE-p3
14.0-RELEASE-p3
14.0-RELEASE-p3
```

**[MANDATORY] How did you install bastille? (port/pkg/git)**
pkg

**[optional] Steps to reproduce?**
Deploy standard ZFS install of 14.0 Release-p3, no extra configuration. Command sequence fails regardless of running as `root` or admin user with full `sudo` privs.
```
» pkg install bastille
Updating FreeBSD-Latest repository catalogue...
FreeBSD-Latest repository is up to date.
All repositories are up to date.
Updating database digests format: 100%
The following 1 package(s) will be affected (of 0 checked):

New packages to be INSTALLED:
        bastille: 0.10.20231125

Number of packages to be installed: 1

43 KiB to be downloaded.

Proceed with this action? [y/N]: y
[1/1] Fetching bastille-0.10.20231125.pkg: 100%   43 KiB  43.9kB/s    00:01
Checking integrity... done (0 conflicting)
[1/1] Installing bastille-0.10.20231125...
[1/1] Extracting bastille-0.10.20231125: 100%

» pkg info bastille
bastille-0.10.20231125
Name           : bastille
Version        : 0.10.20231125
Installed on   : Fri Dec 22 10:54:34 2023 PST
Origin         : sysutils/bastille
Architecture   : FreeBSD:14:*
Prefix         : /usr/local
Categories     : sysutils
Licenses       : BSD3CLAUSE
Maintainer     : christer.edwards@gmail.com
WWW            : https://bastillebsd.org
Comment        : Jail automation framework
Annotations    :
        build_timestamp: 2023-12-14T22:17:29+0000
        built_by       : poudriere-git-3.3.0-1258-gbc38e2e6
        port_checkout_unclean: no
        port_git_hash  : 08ed7d0165
        ports_top_checkout_unclean: no
        ports_top_git_hash: bb504ccec5
        repo_type      : binary
        repository     : FreeBSD-Latest
[...]

PR #650, opened 2023-12-25
diff patches applied to resolve errors on setup.sh and bootstrap.sh

em-winterschon:
Ref: https://github.com/BastilleBSD/bastille/issues/647


yaazkal:
Thank you @em-winterschon for this PR. I've left some commets in the code for you.

PR #665, opened 2024-01-18
restore check_fib() call

andrewhotlab:
This is needed to restore the behavior which respects the "exec.fib" parameter in bastille "console" command, which was removed by commit b997be5


yaazkal:
Thank you @andrewhotlab, merging now.

PR #666, opened 2024-01-22
rename: adjust all mount points

### LINKED ISSUE #659: [BUG] rename doesn't consider custom MOUNT
The `rename` command contains a `update_fstab` routine to update the fstab file of the jail to the new name. However, this is not applied to custom mounts, e.g., those initiated by a `MOUNT` command in a template.

0.10.20231125
14.0-RELEASE-p3
14.0-RELEASE-p3
14.0-RELEASE-p4

pkg

1. create a jail named `foo`
2. apply a template that contains a `MOUNT` command, e.g., `MOUNT /root/bastille/data data nullfs rw 0 0`
3. rename the jail: `bastille rename foo bar`
4. see how the custom entry in the jail fstab file still points to the old path: `/root/bastille/data /usr/local/bastille/jails/foo/root/data nullfs rw 0 0`

The `rename` command takes care of updating *all* entries in fstab, so my entry would be updated to `/root/bastille/data /usr/local/bastille/jails/bar/root/data nullfs rw 0 0`

gahr:
Fixes #659

### DIFF
--- a/usr/local/share/bastille/rename.sh
+++ b/usr/local/share/bastille/rename.sh
@@ -81,24 +81,7 @@ update_fstab() {
     # Update fstab to use the new name
     FSTAB_CONFIG="${bastille_jailsdir}/${NEWNAME}/fstab"
     if [ -f "${FSTAB_CONFIG}" ]; then
-        # Skip if fstab is empty, e.g newly created thick or clone jails
-        if [ -s "${FSTAB_CONFIG}" ]; then
-            FSTAB_RELEASE=$(grep -owE '([1-9]{2,2})\.[0-9](-RELEASE|-RC[1-9])|([0-9]{1,2}-stable-build-[0-9]{1,3})|(current-build)-([0-9]{1,3})|(current-BUILD-LATEST)|([0-9]{1,2}-stable-BUILD-LATEST)|(current-BUILD-LATEST)' "${FSTAB_CONFIG}")
-            FSTAB_CURRENT=$(grep -w ".*/releases/.*/jails/${TARGET}/root/.bastille" "${FSTAB_CONFIG}")
-            FSTAB_NEWCONF="${bastille_releasesdir}/${FSTAB_RELEASE} ${bastille_jailsdir}/${NEWNAME}/root/.bastille nullfs ro 0 0"
-            if [ -n "${FSTAB_CURRENT}" ] && [ -n "${FSTAB_NEWCONF}" ]; then
-                # If both variables are set, update as needed
-                if ! grep -qw "${bastille_releasesdir}/${FSTAB_RELEASE}.*${bastille_jailsdir}/${NEWNAME}/root/.bastille" "${FSTAB_CONFIG}"; then
-                    sed -i '' "s|${FSTAB_CURRENT}|${FSTAB_NEWCONF}|" "${FSTAB_CONFIG}"
-                fi
-            fi
-
-            # Update linuxjail fstab name entries
-            # Search for either linprocfs/linsysfs, if true assume is a linux jail
-            if grep -qwE "linprocfs|linsysfs" "${FSTAB_CONFIG}"; then
-                sed -i '' "s|.${bastille_jailsdir}/${TARGET}/|${bastille_jailsdir}/${NEWNAME}/|" "${FSTAB_CONFIG}"
-            fi
-        fi
+        sed -i '' "s|${bastille_jailsdir}/${TARGET}|${bastille_jailsdir}/${NEWNAME}|g" "${FSTAB_CONFIG}"
     fi
 }
 


yaazkal:
Thanks @gahr for the PR. We had an issue (#698) related to devfs rules that is fixed now. Maybe that could be the root cause. We would appreciate it if you could test again to check if this PR is still needed.

gahr:
Hi @yaazkal , how is this related to devfs rules? This is about mount point mounted via the MOUNT command in Bastillefiles that are not in fstab.

yaazkal:
Sorry @gahr that comment was intented for another PR which is now merged (multi tab in browser is bad for me). Let me take a look at your PR.

yaazkal:
Let me ask, did you test your PR with linux jails too?

gahr:
No, I don't have any Linux jails

gahr:
@yaazkal why do you think this would be jail-type specific? This is the host's fstab, not the guest's.

tschettervictor:
Can we go over this again? I have many jails that I do renames on, and this is a problem for them.

usr/local/share/bastille/rename.sh
```diff
@@ -81,24 +81,7 @@ update_fstab() {
     # Update fstab to use the new name
     FSTAB_CONFIG="${bastille_jailsdir}/${NEWNAME}/fstab"
     if [ -f "${FSTAB_CONFIG}" ]; then
-        # Skip if fstab is empty, e.g newly created thick or clone jails
```
tschettervictor, line 84: Why are you removing all these lines?

gahr: I don't have much familiarity with this code, I just fixed it to fit my use case, so let's go through it.

My proposal is to change occurrences of `${bastille_jailsdir}/${TARGET}` into `${bastille_jailsdir}/${NEWNAME}` in the fstab file. I think this is hardly debatable: we are renaming the jail `${TARGET}` into the jail `${NEWNAME}`, so all mount points that were pointing to the former now need to point to the latter.

So why did I remove all other lines? I think the old code was trying to match a particular mount point into fstab, namely, the `..../root/.bastille` mountpoint. The original bug is that it left all other custom mount points alone, unrenamed.

So, my take is to sed them all, which makes the rest of the code useless.

The linprocfs/linsysfs special case should also be covered, although I don't understand why the leading dot in the string matched in `sed -i '' "s|.${bastille_jailsdir}/${TARGET}/|${bastille_jailsdir}/${NEWNAME}/|" "${FSTAB_CONFIG}"`.

Hope this helps clarify.

tschettervictor: Yes I see what you mean. 
However, what happens if a jail is named the same as a directory in the path specified?

Say you have a directory mounted into a jail named test, and the jail is also named test?

What do you think would be the best way to solve this?

gahr: How would that match `${bastille_jailsdir}/${TARGET}` or `${bastille_jailsdir}/${NEWNAME}`? You'd have to have a path, say, /usr/local/bastille/jails/foo *under* `bastlle_jailsdir`, so your mount point could be:


```
/host/path /usr/local/bastille/jails/TARGET/usr/local/bastille/jails/TARGET nullfs ro 0 0
```

In this case, yes, my approach would rename both TARGET to NEW, while probably you only want to rename the first.

I'd say it's still safer than what we're doing now...

tschettervictor: Ahhh yes, I see what you mean. Yes that would definitely be best in this particular case.

gahr: I have no idea.. but I think this file is sourced when you do `bastille rename`

tschettervictor: It looks like the only place it's called is here. I'm good with this change, unless @cedwards has a good reason to retain the original code.

bmac2: @tschettervictor did you test this code on your system also??

tschettervictor: Still working on that...

tschettervictor: My testing confirms that the old code does not rename the mount points at all.
However, I would keep the old code.

In the "clone.sh" function this line
`sed -i '' "s|${bastille_jailsdir}/${TARGET}/root/|${bastille_jailsdir}/${NEWNAME}/root/|" "${FSTAB_CONFIG}"`
is added, which is not present in the "rename.sh"

I would argue that we should just make the "rename" code identical to the "clone" code as they attempt to do the exact same thing.
The current code does in fact change the RELEASE mount point, but does not touch any other custom mount points.

Here is the code from the "clone.sh"

```
 update_fstab() {
    # Update fstab to use the new name
    FSTAB_CONFIG="${bastille_jailsdir}/${NEWNAME}/fstab"
    if [ -f "${FSTAB_CONFIG}" ]; then
        FSTAB_RELEASE=$(grep -owE '([1-9]{2,2})\.[0-9](-RELEASE|-RELEASE-i386|-RC[1-9]|-BETA[1-9]|-CURRENT)|([0-9]{1,2}(-stable-build-[0-9]{1,3}|-stable-LAST))|(current-build)-([0-9]{1,3})|(current-BUILD-LATEST)|([0-9]{1,2}-stable-BUILD-LATEST)' "${FSTAB_CONFIG}" | uniq)
        FSTAB_CURRENT=$(grep -w ".*/releases/.*/jails/${TARGET}/root/.bastille" "${FSTAB_CONFIG}")
        FSTAB_NEWCONF="${bastille_releasesdir}/${FSTAB_RELEASE} ${bastille_jailsdir}/${NEWNAME}/root/.bastille nullfs ro 0 0"
        if [ -n "${FSTAB_CURRENT}" ] && [ -n "${FSTAB_NEWCONF}" ]; then
            # If both variables are set, update as needed
            if ! grep -qw "${bastille_releasesdir}/${FSTAB_RELEASE}.*${bastille_jailsdir}/${NEWNAME}/root/.bastille" "${FSTAB_CONFIG}"; then
                sed -i '' "s|${FSTAB_CURRENT}|${FSTAB_NEWCONF}|" "${FSTAB_CONFIG}"
            fi
        fi
        # Update additional fstab paths with new jail path
        sed -i '' "s|${bastille_jailsdir}/${TARGET}/root/|${bastille_jailsdir}/${NEWNAME}/root/|" "${FSTAB_CONFIG}"
    fi
}
```
I would just add the last line here to the rename function, but keep the old code in place just in case.

tschettervictor: At the same time, the original code was trying to do something as simple as rename the entire RELEASE fstab, when it is far easier to just rename all occurrences of the old jail to the new jail.

Unless @cedwards was trying to do something specific with it...

tschettervictor:
I've added PR #757 to also change the "update_fstab" function. Here it is for reference.

It indeed makes more sense to search the fstab file for all old jail paths, and simply rename them to the new jail paths.

```
update_fstab() {
    # Update fstab to use the new name
    FSTAB_CONFIG="${bastille_jailsdir}/${NEWNAME}/fstab"
    if [ -f "${FSTAB_CONFIG}" ]; then
        # Update fstab paths with new jail path
        sed -i '' "s|${bastille_jailsdir}/${TARGET}|${bastille_jailsdir}/${NEWNAME}|" "${FSTAB_CONFIG}"
    fi
}
```

usr/local/share/bastille/rename.sh
bmac2: @tschettervictor @yaazkal @gahr 

so what is the final fix here ??????

tschettervictor: In my opinion this PR is a valid fix.
It simply renames the jail path of any fstab entry. See below

```
/usr/local/bastille/releases/14.1-RELEASE /usr/local/bastille/jails/jail10/root/.bastille nullfs ro 0 0
/mnt/test /usr/local/bastille/jails/jail2/root/mnt/test nullfs ro 0 0
```
It would only touch the "/usr/local/bastille/jails/**jailname**/root/" part of the fstab. I don't think there is much danger of touching anything else.

It seems the original code tries to validate that the RELEASE actually exists. This might be overkill, because if the original jail is present then it is to be assumed the RELEASE is also present.
I'm good with this PR, it seems to do ok with my tests.

I'd retain the old code just in case...

tschettervictor: But just to make sure it catches the right code, I went ahead and added an intentional space in front so it will only match the jail path and never the host path.

```
update_fstab() {
    # Update fstab to use the new name
    FSTAB_CONFIG="${bastille_jailsdir}/${NEWNAME}/fstab"
    if [ -f "${FSTAB_CONFIG}" ]; then
        # Update fstab paths with new jail path
        sed -i '' "s| ${bastille_jailsdir}/${TARGET}/root/| ${bastille_jailsdir}/${NEWNAME}/root/|g" "${FSTAB_CONFIG}"
    fi
}
```
I'd recommend this change here as well.

gahr: Well aaaactually.. if you want the extra check to only match the second field in fstab, you should consider that fields can be separated by spaces or tabs. You version won't rename a tab-separate fstab entry.

tschettervictor: Good point. Reverted. Thanks.

I tested it and you are correct.

tschettervictor: I think we can go ahead and merge this. The recent "clone.sh" did exactly the same thing, and it is working.

tschettervictor (approved):

tschettervictor:
@yaazkal I just tested with a bionic Linux jail. Rename updated the mount point perfectly.

yaazkal:
Thanks @gahr and @tschettervictor

Merging this. I'll open a new PR moving `update_fstab` from `clone`, `rename` and `import` to common to avoid this kind of issues in the future and have consistency.

PR #683, opened 2024-02-21
rcp: fix -q option handling

robertek:
Using the `-q` option for rcp does not work.

Testing done after the fix:

```
❯ bastille rcp adguard /etc/hosts .

[adguard]:
/usr/local/bastille/jails/adguard/root//etc/hosts -> ./hosts


❯ bastille rcp adguard -q /etc/hosts .

[adguard]:

```

### DIFF
--- a/usr/local/share/bastille/rcp.sh
+++ b/usr/local/share/bastille/rcp.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Copyright (c) 2018-2022, Christer Edwards <christer.edwards@gmail.com>
+# Copyright (c) 2018-2024, Christer Edwards <christer.edwards@gmail.com>
 # All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


yaazkal:
Hi @robertek after testing your patch I got:

```
[-q]: Not found.
```

That was the original error? Beacause I get it before and after the patch.
If that was, the thing is more related to the command not picking the jail correctly. Let me know if you want to try again.

I'v tested on FreeBSD 14.0-RELEASE with Bastille HEAD at [cb5697a](https://github.com/BastilleBSD/bastille/commit/cb5697acdd41a2a6e66130ca0873fd7382a0f20f)

yaazkal:
Thanks @robertek for the report, This was also present in the `cp` command so I've made a fix where the root cause was. Closing this PR.

PR #684, opened 2024-02-22
A couple of RDR enhancements

robertek:
This PR adds two enhancements:

1) RDR list and RDR clear for persistent rules file `rdr.conf`.
The major rationale behind this was, that the `rdr clear` does not clear the `rdr.conf` and only clears the applied pf rules. I didn't want to change the current default behavior (it may be useful in some cases).  

2) RDR ip and dev commands. It allows changing the globally configured ext_if device and defining specific host target ip address. Even this change extends the rdr.conf format it is consistent with the current format and no migration is needed.

Now a following format of `rdr.conf` is allowed:
```
tcp 22 22
udp 53 53 log (all, to pflog1)
ip 192.168.1.1 tcp 22 22
dev wg0 tcp 22 22
ip 192.168.1.1 dev wg0 udp 53 53
ip 192.168.1.1 dev wg0 tcp 23 22 log
```
It allows bigger flexibility of redirection and fix some issues. 
For example when the system is a gateway and the ext_if is the device being a gateway, the former redirection (from any to any) cause a redirect of all the packets heading to that port on this device even if the destination ip was different.
Another possible example of use case is having a two outgoing interfaces (in all my cases an ethernet port and a wireguard tunnel) with a possibility to selectively redirect from.

Then there are two extra commits. One is for the overall rdr.sh cleanup (deduplication, consistent formating, and generalization). And the last commit is update to the rdr documentation, which was not updated for a while.

### DIFF
--- a/docs/chapters/subcommands/rdr.rst
+++ b/docs/chapters/subcommands/rdr.rst
@@ -7,6 +7,9 @@ without modifying pf.conf (assuming you are using the `bastille0` interface
 for a private network and have enabled `rdr-anchor 'rdr/*'` in /etc/pf.conf
 as described in the Networking section).
 
+When using `dev` modifier you need to make sure the proper nat configuration is
+active for specified network device (`nat on DEV from <jails> to any -> (DEV:0)`).
+
 Note: you need to be careful if host services are configured to run
 on all interfaces as this will include the jail interface - you should
 specify the interface they run on in rc.conf (or other config files)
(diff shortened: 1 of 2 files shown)


tschettervictor:
#765 
I was not aware of this PR, so I rewrote most of the rdr script. 
Could you check it out in the above PR to see if would suit you better?
@robertek

tschettervictor:
@robertek

bmac2:
#765 does these improvments

PR #685, opened 2024-02-27
Update zfs-support.rst

tschettervictor:
Update instruction for usecases that use a pool other than zpool

### DIFF
--- a/docs/chapters/zfs-support.rst
+++ b/docs/chapters/zfs-support.rst
@@ -14,7 +14,6 @@ Two values are required for Bastille to use ZFS. The default values in the ``bas
   bastille_zfs_enable=""                                  ## default: ""
   bastille_zfs_zpool=""                                   ## default: ""
   bastille_zfs_prefix="bastille"                          ## default: "${bastille_zfs_zpool}/bastille"
-  bastille_prefix="/bastille"                             ## default: "/usr/local/bastille". ${bastille_zfs_prefix} gets mounted here
   bastille_zfs_options="-o compress=lz4 -o atime=off"     ## default: "-o compress=lz4 -o atime=off"
 
 Example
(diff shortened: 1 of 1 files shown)


docs/chapters/zfs-support.rst
```diff
@@ -14,7 +14,6 @@ Two values are required for Bastille to use ZFS. The default values in the ``bas
   bastille_zfs_enable=""                                  ## default: ""
   bastille_zfs_zpool=""                                   ## default: ""
   bastille_zfs_prefix="bastille"                          ## default: "${bastille_zfs_zpool}/bastille"
-  bastille_prefix="/bastille"                             ## default: "/usr/local/bastille". ${bastille_zfs_prefix} gets mounted here
```
yaazkal, line 17: This is OK since it needs to be on par with the Bastille config sample.

docs/chapters/zfs-support.rst
```diff
@@ -26,3 +25,8 @@ Example
 
 Replace ``ZPOOL_NAME`` with the zpool you want Bastille to use. Tip: ``zpool list`` and ``zpool status`` will help. 
 If you get 'no pools available' you are likely not using ZFS and can safely ignore these settings.
+
+If you are going to use a zpool other than zroot (eg. ``tank``)  you must also edit the ``bastille_prefix`` (at the top of the ``bastille.conf`` file) to specify the path to the zpool.
+
+Example
+If you choose to run bastille on a pool you have created called ``tank`` you must set ``bastille_prefix=/tank/bastille`` otherwise it will try to use the default ``/usr/local/bastille`` prefix, causing issues with trying to create your containers on your speicified zpool.
```
yaazkal, line 32: Why we should add this since there is the `bastille_zfs_zpool` configuration option?

That said, the `bastille.conf.sample` needs a patch here: https://github.com/BastilleBSD/bastille/blob/c7f46c3fbba3aa5801589fcb0658800a14f71c36/usr/local/etc/bastille/bastille.conf.sample#L44

Note that it says a default value in the comment, but it's not used in the configuration option, so that should be root cause of the described behavior in your PR, could be a bug rather than doing what you describe.

tschettervictor:
Did you test? From my experience the very first variable “bastille_prefix” is set to “/usr/local/bastille” and will create the jails there regardless of the “zfs” settings.

Issue #678, opened 2024-02-10
Some jails not starting on boot


dkaparis:
Some of my jails do not automatically start after boot, but stay down. Manually starting them with `bastille start` works without issue. I have not set `bastille_list` or `bastille_rcorder` and none of my jails have `KEYWORD: nostart` or such in configuration.
Setting `bastille_list=ALL` or explicitly listing all jail names does not affect the issue - the same jails remain down after boot until manually started.

**[MANDATORY] Bastille and FreeBSD version (paste ``bastille -v && freebsd-version -kru`` output)**

```
0.10.20231125
13.2-RELEASE-p8
13.2-RELEASE-p8
13.2-RELEASE-p9
```

All jails are 13.2-RELEASE-p9.

Bastille installed through pkg.

yaazkal:
Thanks, this has been fixed now. Do you want to try again installing Bastille from git?

dkaparis:
It turned out my jails were failing to start on boot because they were trying to mount filesystems that were not available yet. Re-arranging fstab's solved the issue for me. Probably this can be closed.

yaazkal:
ok, thanks for the feedback

PR #686, opened 2024-02-29
[BUGFIX] Fixed rcorder startup

deadbeef2000:
Adding `bastille_rcorder="YES"` to `/etc/rc.conf` doesn't start all jails on bootup. The culprit seems to be `xargs basename`. Without the `-a` flag, this command will only output one match (the first one found).

This fix hasn't been thoroughly tested, although it works in all of my personal usecases.

This should also fix issue https://github.com/BastilleBSD/bastille/issues/678


yaazkal:
Thank you @deadbeef2000 merging now.

PR #689, opened 2024-03-13
Fix to correctly check ZFS on/off

web-sst:
This PR corrects the test for argument compatibility with zfs when exporting jails.


yaazkal:
Thanks @web-sst this is now merged

PR #697, opened 2024-05-15
ZFS check for export.sh need to be inverted

hackacad:
The checkyesno for zfs needs to be inverted. 
`checkyesno` returns 0 if true; see `common.sh:121`
if zfs_enabled=NO , then disallow GZIP etc.


KiriakosGeorgiou:
I keep having to apply this manually over and over (new installs :)

yaazkal:
Thanks @hackacad, this is going to be addressed on PR#689 which was previous to this, so marking this as a duplicate and closing it.

PR #702, opened 2024-05-31
Added the ip_hostname option when creating a jail

bramton:
The `ip_hostname` option allows the jail to resolve it's own IP address. This is very useful in cases where you run your own DNS resolver and ensures that IP addresses can be managed at a single location.

The create subcommand documentation has been modified to reflect this change. In addition, the SLAAC and DHCP options are also added.


yaazkal:
Note that this is only going to work if there are good DNS configuration and also if the hostname has been given correctly to the jail, which is not Bastille responsibility; actually, the user should take care if using this on a jail serving DNS. Why include this as a creation option and not let the user establish it using the `config` command?

bramton:
Agreed. The user has full responsibility for running a working DNS solution. I tried with the `config` command, which works, but is a bit more cumbersome. Here is my recipe:
```
bastille create bogus 14.0-RELEASE 1.2.3.4
bastille config bogus set ip_hostname true
bastille config bogus set ip4.addr
```
Goal of the last line is to remove the ip4 address from the config file, but the downside is that the last line boils down to `ip4.addr;` in the jail.conf file. A subsequent  change of the ip4 address will add an additional line to the jail.conf file. E.g. after running `bastille config bogus set ip4.addr 42.42.42.42`, the jail.conf will look like this:
```
ip4.addr;
ip4.addr = 42.42.42.42;
```
The config file is still functional, but not very pretty. I am not sure if there is a way to prevent this, any thoughts?

yaazkal:
OK. I don't want to close this PR since maybe anyone argues for that option at the creation time. Labeling now as "help wanted" if anyone wants to join the conversation. Let's also see what @cedwards thinks about this option. What I don't want is a plethora of issues in the future with the "sorry, it was DNS" as an answer.

That said, you are pointing that the `config` command maybe wants to have an `unset` option, which could be nice improvement. Please open a new PR with that feature if you want.

About the subsequent change of the `ip.addr4`, that could be a bug then.

Thanks for the feedback.

tschettervictor:
This is not working for me. How and where should the ip_hostname value be filled in?

I'm getting "ip_hostname not a valid IP"

tschettervictor:
Never mind. I got it.

I think there is a better way to implement this though.
Let me know if you are interested.

bramton:
I am still interested in this feature. My current approach is to set a bogus ip address and later change the jail config file manually to enable the `ip_hostname` feature.

tschettervictor:
I've integrated in as well as "inherit" in the "new_rewrite" branch.

Let me update it today and I'll report back here. You can test it then.

It also support a new -D switch for "dual" mode which will add it for both IPv4 and IPv6. If not specified, only IPv4 will be chosen.

bmac2:
PR #830 has this functionality in it.  Closing as we  are  going to merge 830

@yaazkal @tschettervictor

PR #711, opened 2024-07-14
[FIX] correctly use bastille_zfs_prefix to also work on pools != zroot

yaazkal:
See PR #685


tschettervictor:
This won’t work. See #713
The issue I was having was when zfs was set up, it would still use the “default path” at the top of the file (/usr/local/bastille) when creating jails.

yaazkal:
@tschettervictor you are right, I'll try to spare some time next weekend to check it deeply as it solved one issue but it's opening a new one oof.

PR #714, opened 2024-08-07
Allow custom releases

frikilax:
### Changed
- [CREATE] Allow to create jails from unknown releases, as long as the release folder exists
- [CLONE] Improve cloning and fstab update to adapt to custom release names

### DIFF
--- a/usr/local/share/bastille/create.sh
+++ b/usr/local/share/bastille/create.sh
@@ -634,6 +634,7 @@ CLONE_JAIL=""
 VNET_JAIL=""
 LINUX_JAIL=""
 STATIC_MAC=""
+VALIDATE_RELEASE="1"
 
 # Handle and parse options
 while [ $# -gt 0 ]; do
@@ -711,6 +712,10 @@ while [ $# -gt 0 ]; do
             VNET_JAIL_BRIDGE="1"
             shift
             ;;
+        --no-validate|no-validate)
+            VALIDATE_RELEASE=""
+            shift
+            ;;
         --*|-*)
             error_notify "Unknown Option."
             usage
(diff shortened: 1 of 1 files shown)


bmac2:
@frikilax  can you explain your use case please?   you are wanting to bypass bootstrapping a release and just manually copy the release into a folder and use it??

bmac2:
@frikilax  need you to resolve the conflicts caused by shellcheck @yaazkal did on the files.  And explain your use case to me so i know how to test this

frikilax:
Hey @bmac2 !
My use case is : using a custom release name, in a manual folder
We have use cases where we currently use Bastille to launch temporary jails to run CI/CD jobs, those jobs are based on a specific branch/version of the OS, so wee suffix the release name with the branch to run the jobs from

The PR's conflict has been resolved, feel free to come back to me if there's any remaining question/problem!

bmac2:
@frikilax  there are more conflicts probably caused by the two PRs I merged today.   Please clear that and  I will work on this one soon.

tschettervictor:
> @bmac2 I see in the conflicts that the `update_fstab()` function in close.sh doesn't update the additional mountpoints in the jail's fstab anymore: is it desired?
> 
> are those other paths still updated somewhere else?

The reason the fstab function was changed is because it is not necessary to grep for all the content regarding releases.

The only necessary thing is to grep for the old jail path, and replace it with the new jail path.

Thoughts?

frikilax:
> > @bmac2 I see in the conflicts that the `update_fstab()` function in close.sh doesn't update the additional mountpoints in the jail's fstab anymore: is it desired?
> > are those other paths still updated somewhere else?
> 
> The reason the fstab function was changed is because it is not necessary to grep for all the content regarding releases.
> 
> The only necessary thing is to grep for the old jail path, and replace it with the new jail path.
> 
> Thoughts?

@tschettervictor you're right, every mountpoint will be prefixed by the same `${bastille_jailsdir}/${JAILNAME}/root/`, so the new `update_fstab()` seems OK!
(we could argue the function won't work if the `bastille_jailsdir` parameter is changed in config, but many other things won't work in that case)
I'm rebasing my work to let you merge the PR!

frikilax:
@bmac2 the conflicts have been resolved, the PR is all yours

bmac2:
Bug found in cloning @tschettervictor is fixing it

@yaazkal @frikilax

PR #720, opened 2024-09-25
Update common.sh

macmattes:
jail.conf VNET config failed

typo corrected


tschettervictor:
Are you sure this is necessary? The quotes are already being escaped, and therefore are treated as literal characters.
What is the error you are facing that caused the VNET config to fail?

yaazkal:
> jail.conf VNET config failed

How? please describe the issue you are trying to solve.

Double check it and let us know

bmac2:
@macmattes  what are you trying to do here??  The code funcitons correctly as is.  Please explain the error/issue you are having.

PR #723, opened 2024-10-08
Add Support for Static MAC Address for Jails

tschettervictor:
This PR adds support for static MAC addresses for jails. Steps outlined below.

1. It will use the first half of the hosts MAC address to prevent network problems.
2. It will hash the jail name and extract the first five characters of the hash, then append an "a" for the host side and a "b" for the jail side epair MAC address.
3. It will add these values to the jail.conf file for the jail, so it will use the same MAC address on a reboot, and if the jail is recreated, providing the same jail name is used.

I have tested this with the "-B" and "-V" options, and my testing was successful.

Edits are welcome if needed.


tschettervictor:
Closed in favor of #725

PR #728, opened 2024-11-05
Update BASTILLE_VERSION string

subnetspider:
Please update the version string in `usr/local/bin/bastille` from `0.10.20231125` to `0.11.20241022` so that the version reported by `bastille -v` matches the version reported by `pkg info bastille | grep Version`.

Currently, this is not the case:

```shell
admin@freebsd:~ % pkg info bastille | grep Version
Version        : 0.11.20241022
```

```shell
admin@freebsd:~ % bastille -v
0.10.20231125
```

### DIFF
--- a/usr/local/bin/bastille
+++ b/usr/local/bin/bastille
@@ -62,7 +62,7 @@ bastille_perms_check() {
 bastille_perms_check
 
 ## version
-BASTILLE_VERSION="0.10.20231125"
+BASTILLE_VERSION="0.11.20241022"
 
 usage() {
     cat << EOF


foudfou:
Incidentally I noticed the version strings appears in a couple of other files (`docs/chapters/installation.rst` and `docs/conf.py` mainly) and probably need to be updated there as well.

Moreover, and this is probably outside the scope of this PR, I was wondering if/how the process could be automated with github automated releases :thinking:
I'm pretty sure we could add a github workflow to react on new tag, which could at least:

1. replace the version string with the one extracted from the tag
2. commit the new code
3. create a github release

I'm not sure about the integration with readthedocs.

@cedwards @bmac2 have you already given it some thoughts?

foudfou:
After a little research, it should be fairly simple to leverage a github workflow with the github CLI to update the version string and create a new release.

yaazkal:
Yes, it's true that some github actions could help as well as a script. Another PR in one way or another will help (I'll prefer a script).

Anyway some kind of automation is needed.

yaazkal (approved):
thanks

PR #729, opened 2024-11-05
fix default bastille_zfs_prefix in bastille.conf.sample

msoudny:
Remove bastille_zfs_zpool from bastille_zfs_prefix in bastille config.


usr/local/share/bastille/setup.sh
```diff
@@ -30,24 +30,28 @@
 
 bastille_config="/usr/local/etc/bastille/bastille.conf"
 . /usr/local/share/bastille/common.sh
+
+if [ ! -f ${bastille_config} ]; then
+    cp /usr/local/etc/bastille/bastille.conf.sample ${bastille_config}
+fi
```
yaazkal, line 36: This actually has no effect, please use "" when checking the variable

like `if [ ! -f "${bastille_config}" ]; then`

usr/local/share/bastille/setup.sh
```diff
@@ -30,24 +30,28 @@
...
 . ${bastille_config}
 
 usage() {
-    error_exit "Usage: bastille setup [pf|bastille0|zfs|vnet]"
+    error_exit "Usage: bastille setup [pf|network|zfs|vnet]"
```
yaazkal, line 40: I see you are trying to make a change here that is not related to the topic or title of the PR.

Anyway, after executing the `setup` command, it actually created the `bastille0` bridge, so if you want this change to be here, please check that everything has been changed including the documentation since it introduces a break change.

I'll suggest keeping this PR for the bastille_zfs_prefix thing and then opening a new PR for the networking topic as it needs more work.

yaazkal (commented):
Thanks for the PR, please check my comments.

usr/local/share/bastille/setup.sh
msoudny: fixed, thank you

usr/local/share/bastille/setup.sh
msoudny: bastille0 is the default - predefined name, so I do not see issues with it. Will have a look at other points (docu).
But yes, let's remove this from the zfs topic and I will create another PR.

msoudny: or I could simply (for now and backwards compatibility) "add" trhe bastille0 option into the setup.sh

yaazkal: ok, make a decision in order to be able to merge it ASAP since this zfs option is a silly issue but will produce many reports after it hits ports. If you ask me, I'll prefer another PR.

I'll like to request to @cedwards a release just for that after this is fixed.

yaazkal:
This merge is wrong, why deleting `setup.sh` @msoudny ?

yaazkal:
Oh I see, you did created another PR, ok

msoudny:
I hope this PR is canceled. If not, get rid off of it please.

yaazkal:
> I hope this PR is canceled. If not, get rid off of it please.

Yes, is closed, do not worry

PR #733, opened 2024-11-10
Github release on version change

foudfou:
Draft for a more automated release process. May help #728.

The proposed approach is: on git changes to `usr/local/share/bastille/version.sh`, add a new release tag and create a github release.


usr/local/share/bastille/version.sh
```diff
@@ -0,0 +1 @@
+BASTILLE_VERSION="0.11.20241023"
```
foudfou, line 1: Could also add a comment there to explain that changes trigger a CI release flow.

docs/conf.py
```diff
@@ -1,9 +1,9 @@
 import os
 on_rtd = os.environ.get('READTHEDOCS') == 'True'
 if on_rtd:
-    html_theme = 'default'
-else:
     html_theme = 'sphinx_rtd_theme'
+else:
```
foudfou, line 5: Not 100% sure of the original intent here, but `sphinx_rtd_theme` seems more consistent.

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
+name: release
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'master'
+    paths:
+      - 'usr/local/share/bastille/version.sh'
```
yaazkal, line 9: Seems not the perfect place to live, since Bastille expects actual commands in that path. Maybe a .version file in the root or in `util/version.sh` if that file will contain any logic.

.github/workflows/release.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  release:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: bastille-version
+        shell: bash
```
yaazkal, line 20: `runs-on` could be `macos-latest`? Not sure if that will need to change the shell to `sh`, but is just to honor real unix and freebsd legacy.

Regards :)

.github/workflows/release.yml
foudfou: I see: like if someone issues `bastille version` and it would do unexpected things. But I don't think the reasoning holds because:
1. `common.sh` also lives in `/usr/local/share/bastille` and is not a command
2. the entry point script takes extra care about [what commands are allowed](https://github.com/foudfou/bastille/blob/f8ef0d35bf7fa6c9fe7dffc374a430b5b0985921/usr/local/bin/bastille#L142) and explicitly handles the case of `version`.

I guess the alternative would be to move `common.sh` and `version.sh` elsewhere. But I'm unsure where actually. For the sake of simplicity I would thus keep them in the main bastille lib dir.

.github/workflows/release.yml
foudfou: :facepalm: great catch! That's a tough one.

I'm now hesitant: 

On one hand if this PR is the first step towards more automation, there might be a need for deeper discussion on where/how CI should happen. The obvious solution is to run our own runner, but that's actually [not recommended by GH for public repos](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/adding-self-hosted-runners) and has the disadvantage that someone in the community would have to maintain the runner. We could also explore https://github.com/marketplace/actions/freebsd-vm to still run on GitHub. Also for comparison Poudriere uses Cirrus CI, and incidentally autotools for package versioning.

On the other hand, if it's just for version release automation, this PR works ok as is ( :crossed_fingers: ). I agree `macos-latest` is probably slightly more suited.

.github/workflows/release.yml
yaazkal: I understand your argument about `common.sh` but take in mind that today `bastille version` actually does something, just that the `version.sh` file does not exist since the subcommand is directly handled in the `/usr/local/bin/bastille` file.

That said, in order to touch as little files as possible and not create a confusion for later contributions, I'll still suggest to hold that version info file in another location or file.

The `include` directory is not a bad idea for the `common.sh` file and later for the `version.sh` file if you want, but that will translate in a refactoring, so maybe we can check that later since that will need a lot of testing to ensure nothing broke.

.github/workflows/release.yml
yaazkal: By now this should do the version thing only. We should think in CI and testing later of course for GitHub actions, but at this point if it does the version thing correctly then it will be ok if uses `macos-latest` if possible.

Anyway we will need to discuss this with @cedwards and test it before doing any merge.

PR #737, opened 2024-11-23
mount.sh now accepts nullfs mounted files

nullobsi:
It is sometimes desirable to nullfs mount a single file into a jail, like resolv.conf, hosts, or otherwise. However, using a MOUNT in a Bastillefile with a file mount currently throws an error. The script needs a small adaptation to accept files.

### DIFF
--- a/usr/local/share/bastille/mount.sh
+++ b/usr/local/share/bastille/mount.sh
@@ -70,7 +70,7 @@ fi
 ## if host path doesn't exist, type is not "nullfs" or are using advanced mount type "tmpfs,linprocfs,linsysfs, fdescfs, procfs"	
 if [ "${_hostpath}" == "tmpfs" -a "$_type" == "tmpfs" ] || [ "${_hostpath}" == "linprocfs" -a "${_type}" == "linprocfs" ] || [ "${_hostpath}" == "linsysfs" -a "${_type}" == "linsysfs" ] || [ "${_hostpath}" == "proc" -a "${_type}" == "procfs" ] || [ "${_hostpath}" == "fdesc" -a "${_type}" == "fdescfs" ]  ;  then
     warn "Detected advanced mount type ${_hostpath}"
-elif [ ! -d "${_hostpath}" ] || [ "${_type}" != "nullfs" ]; then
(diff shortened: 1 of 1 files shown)


bmac2:
PR and reason makes sense.  We need to test this one and then merge if no issues.
@yaazkal @cedwards

bmac2:
@tschettervictor  can you test this one also so we can make sure it is good

tschettervictor:
So the flow here seems to be this.

If the host path is a directory or a file, then it is acceptable.

I had an issue mounting the file though. This was because the mount command expects the file to be present or it will fail to mount.

I would change the command to the following.

```
    ## Create mount point if it does not exist. -- cwells
    if [ -d "${_hostpath}" ] && [ ! -d "${_fullpath}" ]; then
        if ! mkdir -p "${_fullpath}"; then
            error_exit "Failed to create mount point inside jail."
        fi
    elif [ -f "${_hostpath}" ] && [ ! -f "${_fullpath}" ]; then
        if ! touch "${_fullpath}"; then
            error_exit "Failed to create mount point inside jail."
        fi
    fi
```
This will 
1. Create a directory if the specified path is actually a directory.
2. Simply touch and create the file at the mount point if the specified path is a file and doesn't exist.

The above worked fine for me. @nullobsi

tschettervictor:
I've updated my last comment.

This week ensures that if the fine you are trying to mount will first be created, and also, of there are directories missing in the files path, they will also be created.

Tests are passing with that block.

tschettervictor:
Only things you are missing is inside the "umount" command. That check also needs to be changed from "-d" to "-e"

tschettervictor:
Hmmm. There are actually way more bugs in the mound and unmount that I first realized.
If you don't mind, I'm going to open another PR addressing them.

I'll include this one to allow file mounting also.

bmac2:
@tschettervictor should this one be closed since you opened another PR to address this along with the other mount errors?

tschettervictor:
I believe so. My PR fixes this along with a number of other bugs I found.

bmac2:
fixed in #778

PR #742, opened 2024-11-24
changed setup.sh loopback creation

msoudny:
- check that bastille.conf exists or create a copy from .sample,
- changed "bastille0" loopback creation to read config value $bastille_network_loopback instead,
- changed setup.sh argument "bastille0" for loopback interface creation to a more generic "network",
- "bastille0" as a setup.sh argument is left for backwards compatibility

### DIFF
--- a/usr/local/share/bastille/setup.sh
+++ b/usr/local/share/bastille/setup.sh
@@ -31,10 +31,15 @@
 bastille_config="/usr/local/etc/bastille/bastille.conf"
 . /usr/local/share/bastille/common.sh
 # shellcheck source=/usr/local/etc/bastille/bastille.conf
+
+if [ ! -f "${bastille_config}" ]
+then
+  cp /usr/local/etc/bastille/bastille.conf.sample ${bastille_config}
+fi
 . ${bastille_config}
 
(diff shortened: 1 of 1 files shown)


usr/local/share/bastille/setup.sh
```diff
@@ -31,24 +31,29 @@
 bastille_config="/usr/local/etc/bastille/bastille.conf"
 . /usr/local/share/bastille/common.sh
 # shellcheck source=/usr/local/etc/bastille/bastille.conf
+
+if [ ! -f "${bastille_config}" ]
+then
```
yaazkal, line 36: `if [ ! -f "${bastille_config}" ]; then`

is preferred

usr/local/share/bastille/setup.sh
```diff
@@ -136,8 +141,8 @@ help|-h|--help)
 pf|firewall)
     configure_pf
     ;;
-bastille0|loopback)
-    configure_bastille0
+network|bastille0|loopback)
+    configure_network
```
yaazkal, line 145: Remember to also update documentation

Issue #722, opened 2024-10-01
[BUG] Bastille export is failing


jemail99:
**[MANDATORY] Describe the bug [MANDATORY]**
Using the bastille docs example fails (using my own jail name)
```
bastille export radicale
Error: export option required
```
Also
```
bastille export --xz radicale
Error: export option required
```
However the following work
```
bastille export --tgz radicale
bastille export --txz radicale
```

**[MANDATORY] Bastille and FreeBSD version (paste ``bastille -v && freebsd-version -kru`` output)**
0.10.20231125
14.1-RELEASE-p5
14.1-RELEASE-p5
14.1-RELEASE-p5

**[MANDATORY] How did you install bastille? (port/pkg/git)**
Installed from pkg

**[optional] Expected behavior**
Expect bastille to export the jail using default settings and using --xz option

**[optional]  Additional context**
Running Freebsd on Raspberry pi4 8gb using UFS as the filesystem.

tschettervictor:
Are you using zfs? Looking at the export.sh script it looks like if “bastille_zfs_enable” is not activated it will not even process the “XZ_EXPORT” variable, and will therefore error out.

jemail99:
Hello,
I'm using UFS onlynot ZFS. I assumed that bastille was compatible with UFS and ZFS?

tschettervictor:
It is compatible with UFS. But if you look at the code below, you will see that it will not even register the XZ_EXPORT variable if "bastille_zfs_enable" is not active.

```
jail_export() {
    # Attempt to export the container
    DATE=$(date +%F-%H%M%S)
    if checkyesno bastille_zfs_enable; then
        if [ -n "${bastille_zfs_zpool}" ]; then
            if [ -n "${RAW_EXPORT}" ]; then
                FILE_EXT=""
                export_check

                # Export the raw container recursively and cleanup temporary snapshots
                zfs send ${OPT_ZSEND} "${bastille_zfs_zpool}/${bastille_zfs_prefix}/jails/${TARGET}@bastille_${TARGET}_${DATE}" \
                > "${bastille_backupsdir}/${TARGET}_${DATE}"
                clean_zfs_snap
            elif [ -n "${GZIP_EXPORT}" ]; then
                FILE_EXT=".gz"
                export_check

                # Export the raw container recursively and cleanup temporary snapshots
[...]

PR #744, opened 2024-11-24
export: be clear that --xz is for ZFS systems

yaazkal:
The support for the `--xz` option is only for ZFS at the moment since the `import` command assumes that was the case. By now just show a more descriptive message if `--xz` was used on a UFS system.


yaazkal:
this closes #722

PR #751, opened 2024-11-25
Update usage.rst - bastille list only shows running containers

tschettervictor:
'bastille list' will only print running containers. '-a' or 'all' is needed to print both running and stopped.


bmac2:
list by design ONLY shows running jails.

Approving PR

PR #725, opened 2024-10-09
Support static MAC address + fix clone VNET jail

tschettervictor:
This PR adds some functionality, and some fixes.

- support for static MAC VNET jails (bridged and vnet)
- fix cloning VNET jails
- simplify some sed commands when cloning jails
- allow cloning VNET jails and regenerate MAC address for cloned jail

This PR has been tested with bridged VNET, VNET, and loopback jails. Steps below

1. Create jail
2. Ensure network works
3. Clone jail
4. Ensure network works on cloned jail
5. Ensure jail.conf is updated properly on cloned jail
6. Ensure rc.conf is updated properly on cloned jail

PR #757, opened 2024-12-08
Update clone.sh - simplify "update_fstab" function

tschettervictor:
Due to a discussion in #666 it does not really make sense to include all of this code.

The only things that needs to be changed when cloning or renaming a jail, is the jail path. The code that is being removed make use of a complex set of instructions to replace a simple jail path. 

It is much easier to grep the fstab file for the old jail path, and rename all instances that it finds to the new jail path.


tschettervictor:
Tested and working as expected for me.

bmac2:
your PR for fixing cloning, #725 also changes the clone.sh file.  Is this change in that PR also, or is it a separate change????  can we merge the two so testing cloning can be done once and pushed?  or is there an issue with putting these two together since it is changes to the same file?

@tschettervictor

tschettervictor:
I'll close this and add the changes to the other PR.

PR #758, opened 2024-12-08
Jail autocomplete function - testing required

tschettervictor:
I've rewritten the autocomplete function to be a little easier to follow.
These are the key points.

- main function is inside common.sh
- it is only run if the jail is not first found by grepping the jails directory
- it is only run inside an *.sh file that requires the TARGET=jail arg

When testing, please try to use every single command that is directed toward a jail, and let me know if it works properly or not.

### DIFF
--- a/usr/local/bin/bastille
+++ b/usr/local/bin/bastille
@@ -179,6 +179,12 @@ clone|config|cmd|console|convert|cp|edit|htop|limits|mount|pkg|rcp|rename|servic
               target_all_jails
             else
               JAILS="${TARGET}"
+              # Ensure the target exists
+              if [ ! -d "${bastille_jailsdir}/${TARGET}" ]; then
+                if ! jail_autocomplete "${TARGET}"; then
+                  error_exit "[${TARGET}]: Not found."
+                fi
+              fi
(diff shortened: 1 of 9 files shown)


tschettervictor:
With this current code, the autocomplete function is only called if the jails directory is not detected. 

The function will then proceed to check the directory, and if not found it will exit. If it find multiple directories, it will exist informing the user of multiple choices. If only one is found, it will set it as the TARGET.

tschettervictor:
@bmac2 Not sure where I can view the shellcheck errors. Usually they are listed right inside the file...

tschettervictor:
@bmac2 @yaazkal 

Ive added some shellcheck bypasses for the following reasons.

1. 2010
- I noticed this is bypassed elsewhere, and I can't think of a more practical way to do this. grep is the only way I can find for example "jail" when grepping "ja" so...

2. 2034
- this is because "TARGET" is not used inside "common.sh" but it is still used whenever it is called inside any other scripts.

tschettervictor:
Closing this. There will be a better way to implement this in a future version if PR #776 is accepted.

PR #761, opened 2024-12-08
Update template.sh - bugfix for cmd

tschettervictor:
This fixes a bug where templated "`cmd`" would not working when doing some commands like `fetch`.


bmac2:
@tschettervictor  what was the command syntax that was failing before that now works?  What can I run to see it broken before your change?

bmac2:
fetch command works flawlessly now.  Merging.

tschettervictor:
Sorry. The point of this was that a template that is applied with the CMD argument was failing originally.

bmac2:
tested with template also, works as should

tschettervictor:
@bmac This needs to be reverted. It broke some things. See #400 
Please revert it ASAP.

PR #764, opened 2024-12-09
Update list.sh - bugfix for -p option

tschettervictor:
since the $2 is not passed on to any function by default , we simply pass it to the list_release


bmac2 (approved):

bmac2:
@tschettervictor  please resolve the conficts on this one and I will merge.  I think the shellcheck made a change to that file that wasn't in your branch.  Would prefer you to fix it rather than me mucking with it.
@yaazkal

tschettervictor:
@bmac2 Accidentally closed this. See #773

bmac2:
what release has a patch on it.  I can't find one to test with

PR #767, opened 2024-12-10
Bugfix for mountpoints wrongly creating //

tschettervictor:


tschettervictor:
According to documentation, when mounting or unmounting, the jail path needs to start with a forward slash. "/"
This is standard behavior for mounting and unmounting, but for some reason the actual command required the jail path to not have a leading / which resulted in the fstab file including // whenever a leading slash was provided.

This PR makes sure that the jail path should start with /

tschettervictor:
Ive changed this to retain the old code. This will now simply search for occurrences of // and replace them with /

tschettervictor:
I'm going to close this in favor of #778

Issue #673, opened 2024-01-30
[BUG] PF syntax issues with rdr


jake-g00dwin:
[MANDATORY] Describe the bug [MANDATORY]
Starting/Stoping a thinjail causes pfctl to output an error `DIOCGETRULES: Invalid argument`. When trying to use RDR with ports it will also give errors about a macro not being defined, along with a syntax error.

Sockstat doesn't show the ports being in use before or after trying to start the jails that use RDR.

[MANDATORY] Bastille and FreeBSD version (paste `bastille -v && freebsd-version -kru` output)

```sh
~>$ bastille -v && freebsd-version -kru
0.10.20231125
14.0-RELEASE-p3
14.0-RELEASE-p3
14.0-RELEASE-p4
```

[optional] Steps to reproduce?
```sh
~>$ sudo bastille rdr <jailname> tcp 8448 8448
pfctrl: DIOCGETRULES: Invalid argument
stdin:2: macro '' not defined
stdin:2: syntax error
```

[optional] Expected behavior
Jail up and running using the correct ports.

[optional] Screenshots
![image](https://github.com/BastilleBSD/bastille/assets/134904796/469042f6-3d6c-4d49-a5a3-836fa64dbdf9)

**Bastille List All output**
![image](https://github.com/BastilleBSD/bastille/assets/134904796/8ce3e589-8359-439e-bf9d-852e75cda180)

**PF rules**
![image](https://github.com/BastilleBSD/bastille/assets/134904796/12c4c73f-0d8c-4ad9-97d8-33cfe1938186)

[optional] Additional context

Had a VPS lose power and restart, noticed this issue after that. I'm assuming it's because I had updated Bastille before hand.

jake-g00dwin:
#644 Seems like it could be related, but I haven't changed any system tunables on this system yet, and I don't have IPV6 setup or in use on any interfaces.

adriel-tech:
There were changes to bastille.conf that caused a lot of issues for people on the last Bastille update. Is it possible that you updated Bastille, everything was fine but your VPS rebooting after power loss and reloading the new Bastille settings broke things?

You could try comparing your usr/local/etc/Bastille/bastille.conf with the current one.
https://github.com/BastilleBSD/bastille/blob/master/usr/local/etc/bastille/bastille.conf.sample

jake-g00dwin:
Thanks for the info/fast update. And I think that you're right, that's probably the case.

I've played around with my pf.conf file a bit and manually
[...]

Issue #644, opened 2023-11-30
Syntax error in rdr 


Aquaakuma:
My jail uses the local network loopback method to connect to the Internet, and it works fine. However, when I add any rules using rdr, I get a PF error when I run bastille restart/start jail-name. This error prevents the container from connecting to the Internet, but it recovers after restarting FreeBSD. The error message is as follows: 
```
pfctl:` DIOCGETRULES: Invalid argument
stdin:3: syntax error
pfctl: Syntax error in config file: pf rules not loaded
stdin:4: syntax error
pfctl: Syntax error in config file: pf rules not loaded
```
This is my rc.conf
```
clear_tmp_enable="YES"
sendmail_enable="NONE"
hostname="Amadeus"
#sshd_enable="YES"
inetd_enable="YES"
ntpdate_enable="YES"
ntpd_enable="YES"
powerd_enable="YES"
# Set dumpdev to "AUTO" to enable crash dumps, "NO" to disable
dumpdev="AUTO"
zfs_enable="YES"
cron_enable="YES"


# ipv4 
ifconfig_re0="DHCP"
defaultrouter="YES"

# BastilleBSD
bastille_enable="YES"
cloned_interfaces="lo1"
ifconfig_lo1_name="bastille0"


# Firewall
pf_enable="YES"
pflog_logfile="/var/log/pflog"
pf_rules="/usr/local/etc/pf.conf"
pflog_enable="YES"
gateway_enable="YES"

# SAMBA
samba_server_enable="YES"
```
This is my pf.conf
```
ext_if="re0"

set block-policy return
scrub in on $ext_if all fragment reassemble
set skip on lo

table <jails> persist
nat on $ext_if from <jails> to any -> ($ext_if)
rdr-anchor "rdr/*"

pass in all
pass out quick keep state
antispoof for $ext_if inet
pass in inet proto tcp from any to any port ssh flags S/SA keep state
```
This is my bastille config
```
#####################
## [ BastilleBSD ] ##
#####################

## default paths
bastille_prefix="/usr/local/bastille"                                 ## default: "/usr/local/bastille"
bastille_backupsdir="${bastille_prefix}/backups"                      ## default: "${bastille_prefix}/backups"
bastille_cachedir="${bastille_prefix}/cache"                          ## default: "${bastille_prefix}/cache"
bastille_jailsdir="${bastille_prefix}/jails"                          ## default: "${bastille_prefix}/jails"
[...]

PR #768, opened 2024-12-10
Fix error produced by pfctl when no rules are returned

tschettervictor:
When adding a rdr rule, pfctl will error when it does not find any rules using `pfctl -a rdr/jail -Psn;`

Elsewhere in the script, this is solved by simply adding `2>/dev/null` to not return errors. This PR simply adds the same functionality to the "load_rdr_rule" function.


tschettervictor:
#673 #644

bmac2 (approved):
approve

PR #776, opened 2024-12-16
begin moving functions to common.sh (top htop)

tschettervictor:
@bmac2 

This was a simple first task. Do you think something like this will work?

this basically move the functions to check if a jail is running into common.sh and top and htop are the first files for it to be integrated.

Testing just requires runnin top and htop on a jail. Making sure to see if the command recognized a stopped and started jail. I also added a -f option to start it if it is stopped.

### DIFF
--- a/usr/local/bin/bastille
+++ b/usr/local/bin/bastille
@@ -147,10 +147,10 @@ version|-v|--version)
 help|-h|--help)
     usage
     ;;
-bootstrap|create|destroy|export|import|list|rdr|restart|setup|start|update|upgrade|verify)
+bootstrap|create|destroy|export|htop|import|list|rdr|restart|setup|start|top|update|upgrade|verify)
     # Nothing "extra" to do for these commands. -- cwells
     ;;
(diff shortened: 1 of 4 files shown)


yaazkal:
why accepting `ALL` as a target for `top` and `htop`. How that will work?

tschettervictor:
> why accepting `ALL` as a target for `top` and `htop`. How that will work?

That was already in the top and htop file. It works but you have to press CTRL+C to jump to the next one...

tschettervictor:
I've now changed it to where the set_target_single function will only set the target, and the check_target_exists will only verify that it exists. With this, it is probably necessary to first set, then check existence, the check if TARGET is running (only needed for some commands)

I've also modified it so htop and top now only accept one argument.

tschettervictor:
I've reorganized the functions inside common.sh to alphabetical order.

tschettervictor:
@yaazkal this PR is the first step in migrating some functions.

1. It arranges the commons.sh functions in alphabetical order
2. It moves the "top and "htop" commands to the list of commands that only get passed on inside "/usr/local/bin/bastille"
3. It uses the new "set_target_single" function for both

Let me know if you have any questions.

bmac2:
tested successfully

PR #778, opened 2024-12-24
Mount/Umount fixes and improvements

tschettervictor:
This PR fixes a bunch of bugs found in mount and umount, and adds the ability to mount files as well as directories.

It also included the "set_target" function, as well as a new "error_continue" function that is always called inside loops to continue to the next jail if an error occurs in the event of "ALL" as a target.

To test
1. Mount and unmount directory. Make sure to mount and unmount when they are already mounted to test if it will return that they are already mounted/unmounted
2. Do the above with files also
3. Do the above with directories have spaces in them, making sure to enclose in quotes and use backslashes for the spaces like "bastille mount jail "/my/test\ directory\ here" /mnt/jail/path" or even (not recommended) "bastille mount jail "/my/test\ directory\ here" "/mnt/jail/path\ with\ space""

Make sure that when you retry the mount/umount commands, that it returns the correct error saying that 1. it has been already mounted, 2. the mount doesn't exist

#737

### DIFF
--- a/docs/chapters/subcommands/mount.rst
+++ b/docs/chapters/subcommands/mount.rst
@@ -6,11 +6,40 @@ To mount storage within the container use `bastille mount`.
 
 .. code-block:: shell
 
-  ishmael ~ # bastille mount azkaban /storage/foo /media/foo nullfs ro 0 0
+  ishmael ~ # bastille mount azkaban /storage/foo media/foo nullfs ro 0 0
   [azkaban]:
+  Added: /media/foo /usr/local/bastille/jails/azkaban/root/media/foo
+  ishmael ~ # bastille mount azkaban /storage/bar /media/bar nullfs ro 0 0
+  [azkaban]:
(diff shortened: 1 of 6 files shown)


bmac2:
steps to test this please.

tschettervictor:
Mount a file or folder into a jail, then unmount it.
Try it with a folder and with a file.

It should work.

I was finding that at times it would enter the mount multiple times into the fstab file. So
I moved the grep block to the first thing it does.

Also, this allows to mount files like so "bastille mount jail /etc/rc.conf mnt/etc.conf" or "bastille mount jail /etc/rc.conf /mnt/etc.conf" with or without the leading slash. This PR will trim make sure there are no // in the path.

You can also mount bastille mount jail /etc/rc.conf mnt" and it will  simply mount at /mnt/etc.rc.conf

The old way had bugs like
1. Of the file was not present it couldn't mount.
2. It would overwrite if a file was present, where this PR takes care of that by erroring if a file  is already there

bmac2:
tested working

PR #780, opened 2024-12-29
Add ;;

tschettervictor:


tschettervictor:
@yaazkal Can you test this one? I somehow forgot a ;; at the end of the case command.

bmac2:
I assume this is supposed to be exactly like #770 with the one line fix, but this one is missing other lines that were in 770.  I am confused.  Is the else line needed or not?

tschettervictor:
Yes.

That line is at the end of every case command. I'm not sure why the checks aren't catching it.

tschettervictor:
Out maybe it dude need to be present in a catch all case command.

Are you seeing the same error as @yaazkal ?

bmac2:
no I am just comparing the two PRs and there are other lines in 770.  Specifically a line with:
else
on it missing in this one.  Was that otherline an error too?

bmac2:
the ;; I get and understand

tschettervictor:
Else is missing? Can you point it out?

bmac2:
this one now tests fine for me.  Tested template in normal place, then a template sitting in /tmp for the other test.  Worked through it with @tschettervictor so this one needs another round of testing @yaazkal

yaazkal:
I have tested this and dosen't break anything. Now, I still agree to what @tschettervictor said in the previous PR:

> After careful consideration, it's probably best to leave it as is.
>
> Templates should follow the structure of project/template anyway in my opinion.

If you still want to continue with this PR, it will need documentation so the user dosen't get unexpected behavior or you can consider adding a flag to explicitly tell the system to use the current folder to look for templates. I guess this will be confusing.

tschettervictor:
I'll close it. Let's leave it as is for now.

PR #781, opened 2024-12-29
Fix generate_static_mac

tschettervictor:
I've updated the generate_static_mac function to hash the MAC of the host for the prefix. This is helpful when using interfaces such as bridge0, bridge1 etc.. which all have the same prefix.

This ensures that the MAC will always be unique for each IF/JAIL combo

### DIFF
--- a/usr/local/share/bastille/clone.sh
+++ b/usr/local/share/bastille/clone.sh
@@ -126,8 +126,8 @@ update_jailconf_vnet() {
                     sed -i '' "s|e\([0-9]\{1,\}\)b_${NEWNAME}|e${uniq_epair_bridge}b_${NEWNAME}|g" "${JAIL_CONFIG}"
                     sed -i '' "s|epair\([0-9]\{1,\}\)|epair${uniq_epair_bridge}|g" "${JAIL_CONFIG}"
 		    sed -i '' "s|exec.prestart += \"ifconfig e0a_bastille\([0-9]\{1,\}\).*description.*|exec.prestart += \"ifconfig e0a_${uniq_epair} description \\\\\"vnet host interface for Bastille jail ${NEWNAME}\\\\\"\";|" "${JAIL_CONFIG}"
-                    sed -i '' "s|ether.*:.*:.*:.*:.*:.*a |ether ${macaddr}a |" "${JAIL_CONFIG}"
-                    sed -i '' "s|ether.*:.*:.*:.*:.*:.*b |ether ${macaddr}b |" "${JAIL_CONFIG}"
+                    sed -i '' "s|ether.*:.*:.*:.*:.*:.*a\";|ether ${macaddr}a\";|" "${JAIL_CONFIG}"
+                    sed -i '' "s|ether.*:.*:.*:.*:.*:.*b\";|ether ${macaddr}b\";|" "${JAIL_CONFIG}"
                     break
                 fi
             fi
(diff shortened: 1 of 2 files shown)


bmac2:
@yaazkal this one now tests fine for me.  Test it too please.

@tschettervictor

yaazkal:
I've tested and nothing seems broken. Just for documentation, can you please explain with examples what is the behavior before and after the patch @tschettervictor ?

tschettervictor:
Before the patch there is no option for a jail
To have a statically assigned MAC address.

with the patch, if a user adds the -m option, the jail will be assigned a static mac based on the interface Mac and jail name.

the Mac should always be the same for the same combo of interface/jailname pair per server.

yaazkal:
ok, but why the PR says fix? Is it fixing PR https://github.com/BastilleBSD/bastille/pull/725 ?

tschettervictor:
Oh sorry. 

The old version would generate it but it would not hash the interface MAC meaning that the MAC prefix would always be the same for interfaces like bridge0 bridge1 bridge2 which is unacceptable.

This PR hashes the full MAC of the interface resulting in a different one for every interface.

yaazkal:
Ok, thanks for the clarification. To put it on an example and to understand the logic behind this:

If I have my host interface `vtnet0` and its MAC address is `9a:0d:7c:a3:03:d`; I have a bridge called `vtnet0bridge` and its MAC address is `58:9c:fc:10:9f:07`, what should be the prefix for the jails then?

tschettervictor:
The prefix will be a result of the following code.
The external interface mac also gets sed'd to remove the : character
`echo -n "${external_interface_mac}" | sha256 | cut -b -6 | sed 's/\([0-9a-fA-F][0-9a-fA-F]\)\([0-9a-fA-F][0-9a-fA-F]\)\([0-9a-fA-F]\)/\1:\2:\3/'`

Which result in
vtnet0 = 47:8d:e1 (you left the last character out so i just assumed it is a "d")
vtnet0bridge= 6b:06:0e

yaazkal:
sorry for missing then last character of the vtnet0 MAC, does that change the result?

tschettervictor:
> sorry for missing then last character of the vtnet0 MAC, does that change the result? (I've just edited the missing part)

Yes it will change. if it is a "b" for example, the MAC would then be 14:0a:46

yaazkal:
I'm getting this `77:6f:4c` as a prefix for all the jails. Given that explanation, can't tell then when that value is comming.

tschettervictor:
What is you interface MAC that you are using?

tschettervictor:
That is the correct prefix for the corrected MAC above.

tschettervictor:
```
root@bastille:~ # echo -n 9a0d7ca303d5| sha256 | cut -b -6 | sed 's/\([0-9a-fA-F][0-9a-fA-F]\)\([0-9a-fA-F][0-9a-fA-F]\\)([0-9a-fA-F]\)/\1:\2:\3/'
77:6f:4c
```

tschettervictor:
I think it's fairly random, as each server (if bridges are created) creates them with unique MACs. Also, each servers MAC is unique so the hashed value is also unique.

yaazkal:
> ```
> root@bastille:~ # echo -n 9a0d7ca303d5 | sha256 | cut -b -6 | sed 's/\([0-9a-fA-F][0-9a-fA-F]\)\([0-9a-fA-F][0-9a-fA-F]\\)([0-9a-fA-F]\)/\1:\2:\3/'
> 77:6f:4c
> ```

ok, that makes sense now. Thanks for the clarification. I'll merge.

PR #782, opened 2024-12-29
Improve bastille.conf handling with user prompt for creation

yaazkal:
Replaced ineffective default configuration generation logic with a user-interactive prompt. Users can now choose to create the configuration file with default values if it is missing, ensuring better control and clarity. Removed redundant code from the setup script to streamline execution.


yaazkal:
cc @bmac2

PR #785, opened 2025-01-02
template: Error when multiple spaces in Bastillefile

tschettervictor:
Replace "awk" with "sed" because awk was removing blank lines that were adjacent to other blank lines.

Sed now simply strips the uppercase first arg away from the line, which is what the awk command attempted to do in the first place.


tschettervictor:
#400

PR #787, opened 2025-01-03
template: awk removes spaces from multiple blank lines

tschettervictor:
Awk appears to remove multiple adjacent spaces from lines within a template. Adding "-F '[ ]'" makes sure field splitting is done on every space, thus preserving them.

#400


michael-o (approved):
Yep, and the manpage explicitly says so. WFM now.

PR #793, opened 2025-01-06
config: Correctly match valueless parameters. Fixes #693

### LINKED ISSUE #693: [BUG] CONFIG template hook is NOT idempotent.
CONFIG template hook is NOT idempotent.

0.10.20231125
14.0-RELEASE-p6
14.0-RELEASE-p6
14.0-RELEASE-p6

pkg

- Create a template with any CONFIG directive (ie. `CONFIG set allow.mount;`).
- Run the template more than once (ie. twice).

The jail.conf file should only have 1 entry of `allow.mount;`, but it will have 2 entries instead.

bramton:
The current RE does not match valueless parameters properly. For instance, if the jail config file has a single line with `allow.mount;` the field in the AWK program will have the value `allow.mount;`. This is currently not matched due to the trailing `;`. This PR allows for zero or one trailing `;` in the field.

This resolves #693 and https://github.com/BastilleBSD/bastille/pull/702#issuecomment-2230339823

### DIFF
--- a/usr/local/share/bastille/config.sh
+++ b/usr/local/share/bastille/config.sh
@@ -136,7 +136,7 @@ for _jail in ${JAILS}; do
         awk -F= -v line="${LINE}" -v property="${PROPERTY}" '
             BEGIN {
                 # build RE as string as we can not expand vars in RE literals
-                prop_re = "^[[:space:]]*" property "[[:space:]]*$";
+                prop_re = "^[[:space:]]*" property "[[:space:]]*;?$";
             }
             $1 ~ prop_re && !found {
                 # we already have an entry in the config for this property so


tschettervictor:
What does the "config" command look like before and after this patch?

bramton:
The behaviour of the config command is not changed. To give an example what this PR fixes consider the following fictive jail.conf file:
```
woot {
  host.hostname = woot;
  allow.mount;
  ip6 = enable;
}

```

Now if we run:
```
bastille config woot set allow.mount
bastille config woot set allow.mount
```

The resulting jail.conf now looks like this:

### Before this PR
```
woot {
  host.hostname = woot;
  allow.mount;
  ip6 = enable;
  allow.mount;
  allow.mount;
}
```


### After this PR
```
woot {
  host.hostname = woot;
  allow.mount;
  ip6 = enable;
}
```

tschettervictor:
Ah. I was thinking of the "get" command, which worked ok for me. But you're referring to "set".

I see.
Let me try this.

tschettervictor:
Seems to work, along with setting properties that include values.

bmac2:
worked as advertised.  Merging

louwe:
Thanks for fixing this. May I inquire if this will make it into the next release and when we could expect the version to be released to LATEST package repos?

bmac2:
@louwe you can get it now installing from github.  Hopefully we will have a new release out soon.  Got a couple of more PRs  to get done before next release.

PR #798, opened 2025-01-10
list: correctly print JID instead of JAILNAME

tschettervictor:


tschettervictor:
Removed "" from find commands, as it was erroring with "no such file or directory"

JID prints good now.

bmac2:
tested correctly.  No file not found error, and the JID prints correctly now on bastille list all

yaazkal:
great, thanks

Issue #504, opened 2022-03-04
[BUG] Can't mount tmpfs volume


ChoffaH:
**Describe the bug**
Trying to mount a tmpfs volume fails.
This is what my Bastillefile contains:
```
MOUNT tmpfs tmp tmpfs rw,nosuid,mode=01777 0 0
```

And that results in the following error:
```
Detected invalid host path or incorrect mount type in FSTAB.
Format: /host/path jail/path nullfs ro 0 0
Read: tmpfs tmp tmpfs rw,nosuid,mode=01777 0 0
Failed to execute command: mount
```

**Bastille and FreeBSD version (paste ``bastille -v && freebsd-version -kru`` output)**
```
0.9.20211225
13.0-RELEASE-p4
```

**How did you install bastille? (port/pkg/git)**
pkg

**Steps to reproduce?**
Just add the above config to a jail or run `bastille mount JAIL tmpfs tmp tmpfs rw,nosuid,mode=01777 0 0`

**Expected behavior**
tmpfs volume mounted inside jail

bmac2:
@ChoffaH and @zilti  is this one still an issue??? if so what is the steps so I can reproduce????

ChoffaH:
> @ChoffaH and @zilti is this one still an issue??? if so what is the steps so I can reproduce????

Yes, I can still reproduce this by creating a new jail and running the command under "Steps to reproduce" against it.

bmac2:
@ChoffaH thank you .   Will keep in queue as a bug to be fixed.

bmac2:
https://forums.freebsd.org/threads/mount-tmpfs-inside-jail.47691/

posting on how to mount a tmpfs inside a jail.  Will look at writing up how to do this and add to the docs

tschettervictor:
@ChoffaH can you give #778 a try?

tschettervictor:
Can you test #800?
@ChoffaH

tschettervictor:
Fix merged.

PR #800, opened 2025-01-10
mount: allow mount with options behind permissions

tschettervictor:
Allow mounting with permissions like “rw,other,options” that are needed for tmpfs mounting.

### DIFF
--- a/usr/local/share/bastille/mount.sh
+++ b/usr/local/share/bastille/mount.sh
@@ -89,8 +89,8 @@ elif [ ! -e "${_hostpath}" ] || [ "${_type}" != "nullfs" ]; then
     usage
 fi
 
-# Mount permissions need to be "ro" or "rw"
-if [ "${_perms}" != "ro" ] && [ "${_perms}" != "rw" ]; then
+# Mount permissions,options need to start with "ro" or "rw"
+if ! echo "${_perms}" | grep -Eq 'r[w|o],.*$'; then
     error_notify "Detected invalid mount permissions in FSTAB."
     warn "Format: /host/path /jail/path nullfs ro 0 0"
     warn "Read: ${_fstab}"


yaazkal:
I'll keep the current code. There is no gain in that change and we lost readability.

tschettervictor:
> I'll keep the current code. There is no gain in that change and we lost readability.

Hmm...
We lose a lot without this, namely #504

Current code does not allow mounting with any options other than ro and rw

Thoughts?

yaazkal:
Sorry @tschettervictor I read it wrong. Reopening for testing.

bmac2:
tested good.   This does fix the #504 issue with tmp filesystems.

yaazkal:
My comment to this change is that the user will be forced to always use ro or rw as the first option, is that how it should be? If so let's add that in documentation.

tschettervictor:
You bring a valid point.

You are right. 

We could either do docs, or do a small tweak to make sure rw or ro is always included in the options.

bmac2:
do we need to unmerge this one and make that change??  Force them to use rw or ro?  or is documentation enough?
@yaazkal @tschettervictor

tschettervictor:
I've just submitted PR #801 documenting the behaviour.

yaazkal:
@bmac2 I'll say documentation is enough thinking in the release.

But not sure about if it's preferable by the community to not be force to start wit `ro`, `rw`. Anyway is almost an "standard" to have those as the first option.

But if we think if border scenarios, not sure if anyone is using swap mounts, in that case, that will never work as the option is `sw` or read with quotas wich option is `rq` (Not sure if those also are valid for jails). Check [fstab(5)](https://man.freebsd.org/cgi/man.cgi?query=fstab&apropos=0&sektion=0&manpath=FreeBSD+14.2-RELEASE&arch=default&format=html)

yaazkal:
That said in my previous comment, why we don't make the second option Victor said: check that the basic options are included, in this case as per fstab(5):

>"rw"    /* read/write device */
>"rq"    /* read/write with quotas */
>"ro"    /* read-only device */
>"sw"    /* swap device */
>"xx"    /* ignore totally */

That way, we bring a wide solution rather than an extra limitation to the comment.

What do you think?

bmac2:
fine by me.  @tschettervictor you good with coding that up for this  release?

tschettervictor:
Simple enough. Give me an hour or so to implement it.

tschettervictor:
#802 @yaazkal

PR #801, opened 2025-01-11
mount: document options string to allow comma separated list

tschettervictor:


docs/chapters/subcommands/mount.rst
```diff
@@ -4,6 +4,14 @@ mount
 
 To mount storage within the container use `bastille mount`.
 
+Syntax follows standard `/etc/fstab` format:
+
+.. code-block:: shell
+
+  Usage: bastille mount TARGET HOST_PATH JAIL_PATH [filesystem_type options dump pass_number]
+
+The 'options' string can include a comma-separated list of mount options, but must start with 'ro' or 'rw'.
```
yaazkal, line 13: What do you think about adding an example? It could be an example for tmpfs so that way it illustrates the usage of `ro` or `rw` as the first option

docs/chapters/subcommands/mount.rst
```diff
@@ -17,7 +25,7 @@ Notice the JAIL_PATH format can be /media/foo or simply media/bar. The leading s
 
 It is also possible to mount individual files into a jail as seen below.
 Bastille will not mount if a file is already present at the specified mount point.
-If you do not specify a file name, bastille will mount the file underneath the specified directory as seen in the second example below.
+If the jail file name does not matvh the host file name, bastille will treat the jail path as a directory, and mount the file underneath as seen in the second example below.
```
yaazkal, line 28: typo here in `matvh`

docs/chapters/subcommands/mount.rst
tschettervictor: Done

PR #804, opened 2025-01-11
top/htop: update auto-mode and documentation

tschettervictor:
As per @yaazkal I have update this to now use the new "auto-mode" as opposed to "-f|--force" to auto start the jails if they are stopped.

To test
run htop and top on a running jail to make sure they work.
run htop and top on a stopped jail, using [-a|--auto] to auto-start it.


bmac2:
works as advertised.  starts jails properly when they are down with the -a.   Merging

PR #812, opened 2025-01-14
list: Fix “grep: no such file”

tschettervictor:
This fixes the issue introduced by the 0.12-2025 release.

The problem was the entries were being passed to grep as combined strings because of quotes.

Remove the quotes solves the issue.


bmac2:
Tested and working perfectly now.  Merging.

PR #802, opened 2025-01-11
mount: allow [ro|rw|rq|sw|xx] as available mount options

tschettervictor:
Allow and require at least on of the following strings inside the "options" string of mount.
rw
ro
rq
sw
xx
According to fstab documentation.

Testing
Mount files and directories into jail(s) with different combos of comma separated options, making sure one (and only one) of the above options are also present.

Example `bastille mount jailname /host/path /jail/path nullfs rw,nosuid,mode=01777 0 0` or `bastille mount jailname /host/path /jail/path nullfs nosuid,ro,mode=01777 0 0`

PR #816, opened 2025-01-17
Please fix regex for fstab hook 

MatthiasPetermann:
Commit fb71f0dda55a16e7c26e8f4da055f9896a2f718e introduced to possibility to add options behind permissions
in fstab. Unfortunately it breaks scenarios where no options are provided
as the current regex expects the comma with the options always to be present.

This patch fixes the regex to handle the options as group.

### DIFF
--- a/usr/local/share/bastille/mount.sh
+++ b/usr/local/share/bastille/mount.sh
@@ -92,7 +92,7 @@ elif [ ! -e "${_hostpath}" ] || [ "${_type}" != "nullfs" ]; then
 fi
 
 # Mount permissions,options need to start with "ro" or "rw"
-if ! echo "${_perms}" | grep -Eq 'r[w|o],.*$'; then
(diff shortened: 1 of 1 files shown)


tschettervictor:
There is already a PR addressing this.

#802 

Are you saying you would like to have the option to supply “no options” at all?

MatthiasPetermann:
> There is already a PR addressing this.
> 
> #802
> 
> Are you saying you would like to have the option to supply “no options” at all?

No, I hope I didn’t come across the wrong way. I definitely want to be able to specify the mount permissions (ro, rw, etc.). However, without my patch, the issue is that something like...

```
MOUNT /etc/global /etc/global nullfs ro 0 0
```

produces an validation error.

tschettervictor:
I see.

We are testing the above PR, which only greps to make sure one of [rw|ro|rq|sw|xx] is included according to fstab documentation.

Can you test that?

PR #820, opened 2025-01-19
bridge: Epair name for ^15 characters

tschettervictor:
#786

Allow bridged-VNET interface to work when interface name exceeds 15 characters. This is done by using "epairX" as the name instead of "eXa_jailname" in such cases.

Testing (all with bridged VNET, nothing else should be affected)

1. Create jail called test, rename to test1, validate jail.conf, rc.conf, and networking

2. Create jail called test, rename to testtesttesttest, validate jail.conf, rc.conf, and networking

3. Create jail called testtesttesttest, rename to test1, validate jail.conf, rc.conf, and networking

4. Create jail called testtesttesttest, rename to testing12345678, validate jail.conf, rc.conf, and networking

The purpose of the above is to validate all use cases (short name to short and long name, and long name to short and long name)

5. Do all of the above with "clone" also.

You should find that anytime the new name exceeds 15 characters (plus the interface prefix of eXa_) it will be named "epairXa epairXb" but anytime is is shorter, the current default naming applies.

### DIFF
--- a/usr/local/bin/bastille
+++ b/usr/local/bin/bastille
@@ -164,10 +164,10 @@ version|-v|--version)
 help|-h|--help)
     usage
     ;;
-bootstrap|create|destroy|etcupdate|export|htop|import|list|mount|rdr|restart|setup|start|top|umount|update|upgrade|verify)
+bootstrap|clone|create|destroy|etcupdate|export|htop|import|list|mount|rdr|rename|restart|setup|start|top|umount|update|upgrade|verify)
     # Nothing "extra" to do for these commands. -- cwells
     ;;
-clone|config|cmd|console|convert|cp|edit|limits|pkg|rcp|rename|service|stop|sysrc|tags|template|zfs)
+config|cmd|console|convert|cp|edit|limits|pkg|rcp|service|stop|sysrc|tags|template|zfs)
     # Parse the target and ensure it exists. -- cwells
     if [ $# -eq 0 ]; then # No target was given, so show the command's help. -- cwells
         PARAMS='help'
--- a/usr/local/share/bastille/clone.sh
+++ b/usr/local/share/bastille/clone.sh
@@ -215,6 +215,8 @@ update_jailconf_vnet() {
                     sed -i '' "s|= ${_target_jail_epair};|= ${_new_jail_epair};|g" "${_jail_conf}"
                     sed -i '' "s|up name ${_target_jail_epair}|up name ${_new_jail_epair}|g" "${_jail_conf}"
                     sed -i '' "s|${_target_jail_epair} ether|${_new_jail_epair} ether|g" "${_jail_conf}"
+                    # Replace epair name in jail.conf                  
+                    sed -i '' "s|${_if}|epair${_num}|g" "${_jail_conf}"
                     # If jail had a static MAC, generate one for clone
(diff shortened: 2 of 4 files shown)


bmac2:
ok I tested this one and it made the epair names work as advertised.  I don't use bridged networking so this one I had to follow his steps above.   

@tschettervictor has anyone else tested this beside myself and you?

@yaazkal this one can be tested whenever in the future you get time to test.

tschettervictor:
> ok I tested this one and it made the epair names work as advertised.  I don't use bridged networking so this one I had to follow his steps above.   
> 
> 
> 
> @tschettervictor has anyone else tested this beside myself and you?
> 
> 
> 
> @yaazkal this one can be tested whenever in the future you get time to test.

Not yet. Waiting on @2Bellete to also test.

2Belette:
all working good, I have tested all the testing point listed.
the only thing I noticed, which is probably not due to this PR specifically, is when there is already an epair interface on the system, this is creating an error as there is no test made prior to the creation.
I had and epair2a on my system before creating a long name jail : 

```
bastille create -B -V test2quiesttroplongpour15 14.2-RELEASE 10.0.0.212 vswitch_lan
Valid: (10.0.0.212).
Valid: (vswitch_lan).

Creating a thinjail...

[test2quiesttroplongpour15]:
ifconfig: interface epair2 already exists
jail: test2quiesttroplongpour15: ifconfig epair2 create: failed

[test2quiesttroplongpour15]: Not started. See 'bastille start test2quiesttroplongpour15'.
[test2quiesttroplongpour15]: Not started. See 'bastille start test2quiesttroplongpour15'.
[test2quiesttroplongpour15]:
ifconfig: interface epair2 already exists
jail: test2quiesttroplongpour15: ifconfig epair2 create: failed
```

I don't know if some logic can be tested and if epairX exists, trying to increment X by 1 ?

The rest is all good to me thanks for that

tschettervictor:
There is is a test that exists. I will have to check that out.

tschettervictor:
Can you run the same command with -x to have debug mode enabled?

tschettervictor:
Oh wait, you had the epair already on your system?

The check will only check jail.conf files for the name.

Should we also check the system with ifconfig?

tschettervictor:
@2Belette This seems to be another bug that is unrelated. Basically, bastille will check every jail.conf file and gather a list of defined epairs from there. If you have one that you manually created, bastille will not know about it.

I will put this on my list of things to fix.

2Belette:
@tschettervictor that's exactly right and that the reason I said it is not due to this PR but more something we can improve in the future. So for this PR I think it is all good, but as I had this issue I thought a good idea to report but we may need another Feature Reauest for that

tschettervictor:
It's actually a rather simple fix. I'll update this PR later with the fix included.

tschettervictor:
@2Belette Can you test again?

Make sure to purposely create an epair on your system that is not part of any jails. Then create a jail, you should notice wit will skip over that one and go to the next highest number for the new epair.

2Belette:
Works perfectly ! I even tried to add multiple epair manually and confirmed it is taking the highest and take the next one.
I even tried to add multiple and delete one in the middle manually and it took the one I deleted which was free to use.
Very nice!

PR #829, opened 2025-01-28
Add auto-mode and debug mode to mount and umount

tschettervictor:
This is because if you try to mount and umount while the jail is stopped, you will then run into all sorts of issues when trying to start it because the mount points will already be mounted.

This PR will require the jail to be running when doing mounting and unmounting.

Testing-
Mount and unmount files/directories while jail is stopped. you should find that it will error unless you add "-a" to auto start the jail before attempting the mount.


bmac2:
tested working perfectly

merging

Issue #788, opened 2025-01-04
[ENHANCEMENT] create: Change "interface" to "ip4.addr = if|ip"


tschettervictor:
Since I've started working on PR #783, I've noticed a few minor things that could be tweaked.

1. Classic jails get created in this following format
```
  interface = bastille0;
  ip4.addr = 10.0.0.1;
```

This makes it slightly difficult to deal with adding and removing interfaces from the jail.conf file.

I've currently implemented the following structure when creating jails
`  ip4.addr = bastille0|10.0.0.1`

This makes it really easy to add multiple IP/interface pairs like the following example.
```
  ip4.addr = bastille0|10.0.0.1
  ip4.addr += bastille0|10.0.0.2
  ip4.addr += bastille0|10.0.0.3
  ip4.addr += bridge0|10.0.1.9
```

I will suggest editing create.sh (which I have done easily in my version) to just create classic jails in that way.

bmac2:
I am ok with the change.

@yaazkal your thoughts?

tschettervictor:
@JRGTH

Schueni1:
Oh yes that would be nice.
I am currently defining my network interfaces and ips linke this:
` 
interface = lo0;
ip4.addr = lo0|127.0.0.10/8;
ip4.addr += lagg0|192.168.42.10/24;
`
But having to specify one of the interfaces separately and the rest together with the 'ip4.addr' is a bit weird.

When starting the jail this message is always shown:
`no IP address found for lo0|127.0.0.10/8,lagg0|192.168.42.10/24`
But it seems to work regardless and the ips are getting assigned to the respective interfaces.

tschettervictor:
@Schueni1 You can test #830 which implements this feature.

Keep in mind that you will also need #726 to properly load rdr rules if you use those.

bmac2:
PR 830 merged

PR #765, opened 2024-12-10
RDR Improvements: Allow TO/FROM+INTERACE+IP STACK

tschettervictor:
I've reconstructed much of the rdr.sh to allow users to set TO/FROM and also the interface.
Old command of `bastille rdr jail tcp 8000 80` still functions as it should and will use the default interface and any to any when creating the rules. Major differences are

- you can now set which interface the rule is created on with `-i em0` or any specified interface
- if you want to limit the rdr rule to a certain source, use `-s 134.234.67.34`
- if you have multiple IPs on an interface, you can choose to redirect to only one using `-d 192.168.1.45`
- `-t` is also available if you want to specifically load a rule doing ip4/6 `-t ipv4` or `-t ipv6`. The default will use both, or 'dual'

Error checking is also in place, and testers are welcome.
I will continue to update this PR with the docs and usage commands if users find this helpful.

PR #830, opened 2025-01-29
Allow inherit/ip_hostname as IP options

tschettervictor:
#788 
#369 
#702 
#535 
#521 

Testing

create jail with the below options

inherit as the IP
ip_hostname as the IP
Then do both the above with the "-D|--dual" switch and you should find that it will set them for "ip4" and "ip6"
If you do not ise the -D option, IP6 will be disabled and only ip4 will be set.

Ensure that jails start and stop properly with all the above combinations.

Note: You might need PR #765 if you use RDR at all.


bmac2:
tested on my server with ipv4.  Works as advertised.  I don't have ipv6 on my network so did not test that.

@yaazkal  AFTER 765 this one is ready to test

@tschettervictor

bmac2:
tested extensively with @tschettervictor   
TESTED:
inherit
ip_hostname
checked to make sure pf is not effected negatively by this patch

PR #831, opened 2025-01-29
Fixed rdr rule directive error when a jail has both ipv4 and ipv6 addresses

sinclairday:
When a jail has both an ipv4 and ipv6 address, the PF ``rdr`` directive needs to differentiate between ``inet`` and ``inet6`` address families; otherwise, the firewall rule for the ipv6 interface will fail to load on starting the jail:

```
stdin:3: no translation address with matching address family found.
```

This change changes the address family for ipv6 addresses from ``inet`` to ``inet6``. Specifications for ipv4 addresses are unchanged.

### DIFF
--- a/usr/local/share/bastille/rdr.sh
+++ b/usr/local/share/bastille/rdr.sh
@@ -122,7 +122,7 @@ load_rdr_rule() {
       | pfctl -a "rdr/${JAIL_NAME}" -f-
 if [ -n "$JAIL_IP6" ]; then
   ( pfctl -a "rdr/${JAIL_NAME}" -Psn 2>/dev/null;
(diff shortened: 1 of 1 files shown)


tschettervictor:
Can you test #765?

It has all this and more...

sinclairday:
> Can you test #765?
> 
> It has all this and more...

I did a quick smoke test of #765, and dual-stack worked nicely. I will comment more on the #765 thread.

bmac2:
@sinclairday if you can comment with what you tested on the other PR it would help.  Especially since you are using dual stack 

@tschettervictor @yaazkal

sinclairday:
> @sinclairday if you can comment with what you tested on the other PR it would help. Especially since you are using dual stack
> 
> @tschettervictor @yaazkal

I added a comment to #765 detailing my test and results.

bmac2:
Closing as @sinclairday tested and 765 fixes this issue.  Closing as included in  #765

Issue #841, opened 2025-02-13
[BUG] When bastille is configured to use the loop back interface (bastille0, ip4),  adding a jail with a shared interface does not work


kimATwork:
[MANDATORY] Describe the bug [MANDATORY]

On a bastille host set up with a cloned loop back interface (bastille0, ip4), adding a jail that shares the hosts network interface does not work. The jail is created and is running, and the ifconfig alias is added. But the jail can't make any inbound connections.

The reason is that share/bastille/start.sh adds the jails ip to the pf firewall table regardless of what interface the jail uses.
Also, since share/bastille/stop.sh *do* checks to se if the jails interface matches "bastille_network_loopback", the ip on the shared interface is not removed from the pf firewall table when the jail is stopped.

The problem is solved by adding the appropriate check to start.sh

[MANDATORY] Bastille and FreeBSD version (paste ``bastille -v && freebsd-version -kru`` output)**

#> bastille -v && freebsd-version -kru
0.13.20250126
14.2-RELEASE
14.2-RELEASE
14.2-RELEASE

**[MANDATORY] How did you install bastille? (port/pkg/git)**
pkg

tschettervictor:
Can you test #830 and see if that solves the issue for you?

bmac2:
@kimATwork  test #830 to see if it fixes your issue please.

kimATwork:
@bmac2 @tschettervictor

> [@kimATwork](https://github.com/kimATwork) test [#830](https://github.com/BastilleBSD/bastille/pull/830) to see if it fixes your issue please.

I'm  uncertain about what to look for. It kind of fixes the problem, but introduces another.
In my usecase, #830 creates both jails with new ip4.addr=if|ip format and aparently sets up the interfaces correct. But there where no entries in the pf jail table. With #830  jail_A (if ix1) is reachable but Jail_B (if bastille0) is not since its ip is not insertet into the pf table. So its in fact opposit of v0.13.20250126 where both jails IPs are inserted into the table. The correct is ofc that only jail_Bs ip (bastille0) should be inserted. Which is the target of #842.

Also, in #830 there seems to be a problem with stop.sh, as i can't stop jails with it. It returns usage().

tschettervictor:
The reason you can't use stop is that you need the updated /usr/local/bin/bastille file as well. Did you grab that?

I see your point, so I will integrate it into #830 also.

kimATwork:
[...]

PR #842, opened 2025-02-13
Let jails ip4.addr be added to pf jails table if it isn't local to any interface

kimATwork:
#841
I added the same check that is present in stop.sh, but could just as well be:
`if [ $(bastille config "${_jail}" get interface) == ${bastille_network_loopback} ]; then`

### DIFF
--- a/usr/local/share/bastille/start.sh
+++ b/usr/local/share/bastille/start.sh
@@ -89,7 +89,6 @@ for _jail in ${JAILS}; do
             fi
             ## add ip4.addr to firewall table if its using the loop back interface
             if grep -qw "interface.*=.*${bastille_network_loopback}" "${bastille_jailsdir}/${_jail}/jail.conf"; then
-                info "Adding ${_ip4} to pf table ${bastille_network_pf_table}"
                 pfctl -q -t "${bastille_network_pf_table}" -T add "${_ip4}"
  	    fi
         fi
--- a/usr/local/share/bastille/start.sh
+++ b/usr/local/share/bastille/start.sh
@@ -87,10 +87,14 @@ for _jail in ${JAILS}; do
                 error_notify "Error: IP address (${_ip4}) already in use."
                 continue
             fi
-            ## add ip4.addr to firewall table if its using the loop back interface
-            if grep -qw "interface.*=.*${bastille_network_loopback}" "${bastille_jailsdir}/${_jail}/jail.conf"; then
-                pfctl -q -t "${bastille_network_pf_table}" -T add "${_ip4}"
- 	    fi
+            ## add ip4.addr to firewall table if pfctl is installed, pf table is available and
+            ## the ip4.addr is not reachable trough local interface (assumes NAT/rdr is needed)
+            if [ "${bastille_network_pf_table}X" != "X" ] && \
+                [ -x /sbin/pfctl ] && pfctl -s Tables | grep "${bastille_network_pf_table}" >/dev/null ; then 
+                    if route -n get ${_ip4} | grep "gateway" >/dev/null; then
+                        pfctl -q -t "${bastille_network_pf_table}" -T add "${_ip4}"
+                    fi
+            fi
         fi
 
         ## start the container
--- a/usr/local/share/bastille/start.sh
+++ b/usr/local/share/bastille/start.sh
@@ -87,13 +87,9 @@ for _jail in ${JAILS}; do
                 error_notify "Error: IP address (${_ip4}) already in use."
                 continue
             fi
-            ## add ip4.addr to firewall table if pfctl is installed, pf table is available and
-            ## the ip4.addr is not reachable trough local interface (assumes NAT/rdr is needed)
+            ## add ip4.addr to firewall table if the ip4.addr is not reachable trough local interface (assumes NAT/rdr is needed)
             if route -n get ${_ip4} | grep "gateway" >/dev/null; then
-                if [ "${bastille_network_pf_table}X" != "X" ] && \
-                  [ -x /sbin/pfctl ] && pfctl -s Tables | grep "${bastille_network_pf_table}" >/dev/null ; then 
-                        pfctl -q -t "${bastille_network_pf_table}" -T add "${_ip4}"
-                fi
+                pfctl -q -t "${bastille_network_pf_table}" -T add "${_ip4}"
             fi
         fi
 
--- a/usr/local/share/bastille/start.sh
+++ b/usr/local/share/bastille/start.sh
@@ -87,8 +87,11 @@ for _jail in ${JAILS}; do
                 error_notify "Error: IP address (${_ip4}) already in use."
                 continue
             fi
-            ## add ip4.addr to firewall table
-            pfctl -q -t "${bastille_network_pf_table}" -T add "${_ip4}"
+            ## add ip4.addr to firewall table if its using the loop back interface
+            if grep -qw "interface.*=.*${bastille_network_loopback}" "${bastille_jailsdir}/${_jail}/jail.conf"; then
+                info "Adding ${_ip4} to pf table ${bastille_network_pf_table}"
+                pfctl -q -t "${bastille_network_pf_table}" -T add "${_ip4}"
+ 	    fi
         fi
 
         ## start the container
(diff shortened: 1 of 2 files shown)


tschettervictor:
I don't follow this.

Isn't it necessary to add the jail IP to the firewall table for any interface it is on?

I currently have jails that I connect to the host interface (vtnet0) and they run fine.


Could you give a bit more context.

kimATwork:
> Isn't it necessary to add the jail IP to the firewall table for any interface it is on?
A firewall is not required if the jail uses a shared interface (as in 'ifconfig alias'), since NAT is not needed. In that case the jail has its own IP on the LAN. 
> I currently have jails that I connect to the host interface (vtnet0) and they run fine.
I think VNET is one of three ways of setting up to network according to the docs. If ypu try to combine the two others (cloned interface and alias), you'll hit this bug.
> Could you give a bit more context.

kimATwork:
> Isn't it necessary to add the jail IP to the firewall table for any interface it is on?

A firewall is not required if the jail uses a shared interface (as in 'ifconfig alias'), since NAT is not needed. In that case the jail has its own ip on the same interface/LAN as the host.

> I currently have jails that I connect to the host interface (vtnet0) and they run fine.

Yes, I do to, but I've never run jails on host interface and cloned interface (bastille0) on the same host at the same time, until now.

> Could you give a bit more context.

Sure.

I have a host, bastille01, with ip 192.168.0.10 on interface xl1. After installing bastille I can create a jail like this:
```
[root@bastille01 ~]# bastille create jail_A 14.2-RELEASE 192.168.0.20 xl1
jail_A: created
```
I can reach jail_A by setting up sshd in the jail and ssh to it from any host in the network.

Then i run out of ip adresses so I configure bastille to use a cloned interface (bastille0), activate pf, and sets up the pf.conf to handle NAT and rdr. As the documentation describes.

Now I can create a jail like this:
```
[root@bastille01 ~]# bastille create jail_B 14.2-RELEASE 172.16.2.3 bastille0
jail_B: created
```
To be able to reach sshd in jail_B from the rest of the network, i need to configure port forwarding:

`[root@bastille01 ~]# bastille rdr host_B tcp 22000 22`


Now I can reach both my jails by ssh. When i check the pf table and bastille it looks like this:
```
[root@bastille01 ~]# pfctl -t jails -T show
   172.16.2.3

[root@bastille01 ~]# bastille list -a
 JID  State  IP Address    Published Ports  Hostname  Release           Path
 57   Up     192.168.0.20  -                jail_A    14.2-RELEASE      /usr/local/bastille/jails/jail_B/root
 65   Up     172.16.2.3    tcp/2200:22      jail_B    14.2-RELEASE      /usr/local/bastille/jails/jail_A/root
```
Then, when i restart jail_A, I can no longer reach it via ssh. What happend? If I add port forwarding, it works again, but that should not be necessary. If I check the pf table again, it now look like this:
```
[root@bastille01 ~]# pfctl -t jails -T show
   172.16.2.3
   192.168.0.20
```
Jail_A (192.168.0.20) should not even be there. So I remove it:
`[root@bastille01 ~]# pfctl -q -t jails -T delete 192.168.0.20`

Now, its reachable again.

Then i restart jail_A and it turns unreachable again. And its ip is back in the pf table again.

 OK, so i'll stop both  jails:
```
   [root@bastille01 /usr/local/share/bastille]# bastille stop jail_A
   nat cleared
   [jail_A]:
   jail_A: removed
   [root@bastille01 /usr/local/share/bastille]# bastille stop jail_B
   [jail_B]:
   jail_B: removed
```
Now, jail_Bs ip is gone from the pf table, but jail_As ip is still present:
```
[root@bastille01 ~]# pfctl -t jails -T show
   192.168.0.20
```
So this means that jail_B, thats using the cloned interface, is working as expected but jail_A only works by adding port redirection (which should not be necessary) or by removing Jail_As ip from pf manually after its started. Also, jail_As ip is still present in the pf table even after its stopped.

The code handling pf in start.sh:

```
  83         ## warn if matching configured (but not online) ip4.addr, ignore if there's no ip4.addr entry
  84         _ip4=$(bastille config "${_jail}" get ip4.addr)
  85         if [ "${_ip4}" != "not set" ]; then
  86             if ifconfig | grep -wF "${_ip4}" >/dev/null; then
  87                 error_notify "Error: IP address (${_ip4}) already in use."
  88                 continue
  89             fi
  90             ## add ip4.addr to firewall table
  91             pfctl -q -t "${bastille_network_pf_table}" -T add "${_ip4}"
  92         fi
```

The code handling pf in stop.sh:

```
 77         ## remove (captured above) ip4.addr from firewall table
 78         if [ -n "${bastille_network_loopback}" ] && [ "${_ip4}" != "not set" ]; then
 79             if grep -qw "interface.*=.*${bastille_network_loopback}" "${bastille_jailsdir}/${_jail}/jail.conf"; then
 80                 pfctl -q -t "${bastille_network_pf_table}" -T delete "${_ip4}"
 81             fi
 82         fi
```

stop.sh line 78 checks if bastille_network_loopback is configured and on line 79 it checks that the jails interface matches the loop back interface, before removing the ip from the firewall.

That check is not done in start.sh, which explains why jail_A is put into the pf table, and never removed.
The root problem is that start.sh should do the same test as stop.sh, keeping jail_A out of the pf table.

So that is what my fix aims to do. I've tested it on my own host.

tschettervictor:
So, you're saying that using the alias mode should not have the IP being added to the pf table?
This does make sense, and I'm not sure why the check would exist inside stop.sh and not start.sh
Probably an oversight.

@yaazkal @bmac2

tschettervictor:
Might be slight overkill here.

We should not check for the "loopback address" because it is always filled by default with "bastille0"

Also, the "pf_table" is always defined as jails so no point in checking that either.

We only need to check if the "ip4.addr" is "not set" and grep for the "gateway" when running the "route" command.

kimATwork:
Really? If i wanted to run my own firewall, for example ipfw, i'd just remove the pf related setup from bastille.conf. But then again, without the tests  it will be more visible if the user types in the wrong IP triggering the pf insert,  without having pf installed. I have committed the changes now.

tschettervictor:
There is an open issue requesting support for IPFW.

We are probably going to merge #830 sometime soon, and I've purposely structured it so we can do that.

The issue we have now is

1. How do we tell which is being used? IPFW or PF? (Perhaps another switch inside the conf file? "bastille_network_firewall")

2. How do we tell if the IP is in the local net so as not to add it to the table? (This I think you solve with the route command)

tschettervictor:
Your latest changes look good.

I still struggle to replicate.

Couple questions.

1. Why does the jail work the first time it starts? The IP has already been added to the table, so it doesn't make sense that it would work.

2. How are you testing the outbound connection?

PR #830 which you retested now adds and removes the IPs properly for each jail that is non-VNET, without any extra checks. If you still have issues after that, then we can consider merging this one.

kimATwork:
> 1. How do we tell which is being used? IPFW or PF? (Perhaps another switch inside the conf file? "bastille_network_firewall")

I haven't user IPFW to handle NAT, only stateful filtering and traffic shaping. Pf and IPFW works fine together in my setup. IPFW as a firewall and pf handling NAT. But i think supporting IPFW would be nice. So using "bastille_network_firewall" default set to "No" seems like a good starting point to me. The user will need to configure the firewall of choice anyway.

> 2. How do we tell if the IP is in the local net so as not to add it to the table? (This I think you solve with the route command)

Yes, I think the route command solves it as far as I can see. It should know the hosts network and should always available.

> Why does the jail work the first time it starts? The IP has already been added to the table, so it doesn't make sense that it would work.

It worked initially because I at that time hadn't set up pf. I ran out of IP addresses, and wanted to solve that by using Bastilles 'loopback' option. After configuring pf etc. the second jail worked as expected using 'bastille0'. Then when i restarted the first jail again I could'n ssh into it even if the second jail using loopback interface worked fine. The rest is history i guess :)

> How are you testing the outbound connection?

Fetching index.html from a known Norwegian news paper.

tschettervictor:
Since we are adding this I have added this fix to #830

It should be merged soon.

tschettervictor:
This fix was merged in #830 so closing this.

Issue #858, opened 2025-02-19
[BUG] "Jail not found" when both CLICOLOR & CLICOLOR_FORCE environment variables are set


vegged:
When both shell environment variables **CLICOLOR** & **CLICOLOR_FORCE** are set to 1, several `bastille` commands (`console`, `etcupdate`, `htop`, `jcp`, `rename`, `top`, `update`) fail with the following error message:
```
Jail not found "jailname" 
```

Forced color escape sequences likely mess with pattern matching in `common.sh`, as unsetting **CLICOLOR_FORCE** fixes the issue.

Tested on *Bastille 0.13.20250126* with *zsh 5.9 (amd64-portbld-freebsd14.0)* & *GNU bash, version 5.2.37(0)-release (amd64-portbld-freebsd14.1)* (all installed via `pkg`) on *FreeBSD 14.2-RELEASE-p1*.

vegged:
[This commit](https://github.com/BastilleBSD/bastille/pull/845/commits/1f1cf7238029d304abc2483aaa74f0e073b8df2d) offers a fix.

tschettervictor:
You can run those commands with -x to see what's going on.

bmac2:
so I have to ask.  What is color escape sequences and why do you have  them.  This is not a bastille bug.

vegged:
> so I have to ask. What is color escape sequences and why do you have them. This is not a bastille bug.

Setting the **CLICOLOR** environment variable seems to be a standard method for enabling colorized output for `ls` (as per [ls(1)](https://man.freebsd.org/cgi/man.cgi?query=ls)), while `more -R`, `less -R` require **CLICOLOR_FORCE**.

I believe this isn't an exotic shell environment scenario. Why not accommodate this fairly common FreeBSD setup?

tschettervictor:
Can you share both vars and what you set them too? I'll investigate.

tschettervictor:
This has nothing to do with the common.sh functions, but rather the "list.sh" command that is used inside common.sh to list all the jails.

It uses "ls" to find them, and it looks like that might be interfering if one has the "CLICOLOR_FORCE" set to 1.

tschettervictor:
```
root@dev1:~ # bastille list jails
+ [ 1 -gt 0 ]
+ list_jail
+ [ -d /usr/local/bastille/jails ]
+ ls /usr/local/bastille/jails
+ sed 's/\n//g'
[...]

PR #859, opened 2025-02-20
list: No color when searching for jails

tschettervictor:
This should solve the issue mentioned in #858


bmac2:
was able to reproduce the error consistently.  You have to sudo su to root, NOT su to root to see the bug.  Once this patch was applied the bug was fixed.  Merging.

Issue #815, opened 2025-01-16
[ENHANCEMENT] Bulk start/stop/destroy jails


2Belette:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I am testing Bastille / Template / any other thing outside from Bastille regarding jails where I am creating a bunch of tests jails and have to start/stop or destroy multiple times.

**Describe the solution you'd like**
```
bastille start test1 test2 test3 test4 test5 ...
bastille stop test1 test2 test3 test4 test5 ...
bastille destroy test1 test2 test3 test4 test5 ...
```

**Describe alternatives you've considered**
Doing it manually now

**Additional context**
N/A

yaazkal:
You can use the `ALL` as a target. Like:

```
bastille start ALL
bastille stop ALL
bastille destroy ALL
```

2Belette:
Yes thanks I know that but I don't want to start/stop/destroy all but just the jails test. It would also be very convenient for template to apply in one command for multiple jails but not for ALL

yaazkal:
Ok, so you say supporting something like `bastille CMD test* OPTIONS`?

In the meantime, this is what you can do for testing (change accordingly):

Create 5 jails with names test1, test2, test3, test4 and test5:
`for i in $(seq 1 5); do bastille create test${i} 14.2-RELEASE 10.0.0.${i}; done`

Stop the jails:
`for i in $(seq 1 5); do bastille stop test${i}; done`

Destroy the jails:
`for i in $(seq 1 5); do bastille destroy test${i}; done`

2Belette:
Many thanks for the commands !

bastille test* start would be cool but that's would be only for the test I am doing.

For more maintenance and so on in production, doing a "bastille start|stop|destroy|template jail1 jail2 jail2" would be very convenient

tschettervictor:
I have it local so I can quote jails together.

bastille start "test test1" etc...

2Belette:
> You can use the `ALL` as a target. Like:
> 
> ```
> bastille start ALL
> bastille stop ALL
> bastille destroy ALL
> ```

for information stop ALL is working all fine but when I am trying destroy ALL after stoping them ALL I got : 
```
bastille destroy ALL
Jail not found.
```
I got 11 jails

tschettervictor:
Merged
#830 
#861

2Belette:
Thanks!

PR #861, opened 2025-02-22
destroy: New functions

tschettervictor:
#715 
#815 

Destroy cache on destriying a release unless --no-cache|-c is set
Bulk destroy jails now available by "ALL" or "jail jail2 jail3"

@michael-o 
@2Belette
@joh-ku


bmac2:
tested and deletes the cache correctly.

bmac2:
tested deleting a list of jails and delete ALL.   Works as expected.

PR #870, opened 2025-02-23
template: Implement new functions and redo bastille main exec

tschettervictor:
This PR adds the final touches to finally allowing the main bastille executable to now only redirect commands to their respective scripts. It no more does any functions. Each command now takes care of its own functions like setting the target.

Testing

- run each command to make sure it functions as it should, report any errors here and we will get the corrected


bmac2:
tested

PR #845, opened 2025-02-14
zfs send without -R on bastille create

vegged:
Since there are no descendent file systems in the `bastille create` scenario, stripping `zfs send` of the `--replicate` option enables support for encrypted datasets. Fixes #839.

PR #876, opened 2025-02-25
create: Send without -R for encrypted bastilleroot

tschettervictor:
@vegged
#845

This will check the main bastille dataset to see if encryption is enabled. If it is, it will send without -R.

Can you test?

This will not help or work for cloning though. That will have to be fixed later.

### DIFF
--- a/usr/local/share/bastille/create.sh
+++ b/usr/local/share/bastille/create.sh
@@ -453,6 +453,12 @@ create_jail() {
 
                         ## sane bastille zfs options
                         ZFS_OPTIONS=$(echo ${bastille_zfs_options} | sed 's/-o//g')
+			## send without -R if encryption is enabled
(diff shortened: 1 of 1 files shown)


vegged:
> @vegged #845
> 
> This will check the main bastille dataset to see if encryption is enabled. If it is, it will send without -R.
> 
> Can you test?
> 
> This will not help or work for cloning though. That will have to be fixed later.

Works for me, the previous block caused by `zfs send -R` is bypassed!

On a sidenote, I stil get this, though this does not prevent the creation of the _thick jail_:
```
+ zfs set 'compress=lz4' 'atime=off' 'mountpoint=none' zroot/bastille/jails/testjail/root
cannot unmount '/usr/local/bastille/jails/testjail/root': pool or dataset is busy
+ zfs inherit mountpoint zroot/bastille/jails/testjail/root
cannot unmount '/usr/local/bastille/jails/testjail/root': pool or dataset is busy
```

tschettervictor:
@vegged arr we good to merge this then?

PR #887, opened 2025-02-28
import: Import jails using new interface format (iocage,ezjail)

tschettervictor:
Imported jails from iocage and ezjail should now be imported using the new "if|ip" format and overall better handling on the network side of the import.

Testing requires an exported iocage jail to be imported into bastille.


bmac2:
testing the options for exporting jails and exported multiple jails.  Tested  as expected.

PR #889, opened 2025-03-01
bastille: Allow custom config

tschettervictor:


tschettervictor:
Not working as expected.
config files need to be sourced from each command

Issue #893, opened 2025-03-01
[ENHANCEMENT] --vlan switch


2Belette:
**Is your feature request related to a problem? Please describe.**
Today, there is no automated way to created one (or multiple) VLAN interfaces in a jail created by Bastille.
This becomes even more interesting when the [support of adding multiple interfaces in jails](#783) will be available.

**Describe the solution you'd like**
`--vlan VLANID` during `create` or `network` operations would add a `vnetX.VLANID` to the jail

Example : 
` doas bastille create -B -V test 14.2-RELEASE 10.10.20.21/24 --vlan 20 vswitch_lan`

Would generate a rc.conf with : 
```
ifconfig_vnet1="up" # see my note below
ifconfig_e1b_test_name="vnet1"
vlans_vnet1="20"
ifconfig_vnet1_20=" inet 10.10.20.21/24 "
```
And would create `vnet1.20` interface to the jail

Note : I realized that if I don't explicitly bring the main interface (in my example, vnet1) where the VLAN interface is built from, there is no connectivity on the vnet1.20 interface as the main interface is down. This would need to be tested from someone else to confirm that I am not hitting a one-off issue.

**Additional context**
Something to consider after a large number of testings : I was not able to get connectivity through the VLAN interface, this is not related to Bastille, but just to share my findings as this can occur to others.

I realized that the traffic is coming out from the jail to the bridge and to the external network correctly but the traffic is never coming back to the jail.

I ran tcpdump and external packet capture (with physical probe) from multiple points of measure on my network : ARP Requests are seen and sent over (jail, epair bridge, physical NIC connected to the bridge and to the external network) . ARP Replies are seen/sent from my external inter-VLAN router but never seen by tcpdump ran on the bridge nor the physical interface attached to the bridge (em0 on my side) nor on the epair interface ! BUT seen by a TAP (Test Access Point) connected to an external probe tapping the same em0 interface !

I thought I was originally hitting a known bug on ARP and VLAN/Jails but when I add manually the router MAC/IP address to the jail ARP table the other type of traffic had the same issue (ICMP on my test) ICMP Requests seen all over the paths but the Replies disappeared on my host. 

The kernel is silently deleting all the tagged traffic reaching em0/physical interface on the bridge.
To confirm, I manually created a VLAN interface to the same host bridge and put an IP address and was able to successfully get connectivity.

My host what Bastille documentation is suggesting : 
```
net.link.bridge.pfil_bridge=0
[...]

PR #899, opened 2025-03-03
create: Allow setting VLAN for VNET jails

tschettervictor:
#893 
@2Belette

This PR will allow users to specify a VLAN ID when creating a VNET jail.
You will need to add the bottom line of the update config file to your config file if it does not automatically add it.
This is because a template has been created for VLAN purposes, as that was the best approach.

Usage:

`bastille create -V --vlan 20 jail 14.2-RELEASE 192.168.13.13/24 vtnet0`

Adjust the above to fit your network.

### DIFF
--- a/usr/local/etc/bastille/bastille.conf.sample
+++ b/usr/local/etc/bastille/bastille.conf.sample
@@ -66,3 +66,4 @@ bastille_template_thick="default/thick"                               ## default
 bastille_template_clone="default/clone"                               ## default: "default/clone"
 bastille_template_thin="default/thin"                                 ## default: "default/thin"
 bastille_template_vnet="default/vnet"                                 ## default: "default/vnet"
+bastille_template_vlan="default/vlan"                                 ## default: "default/vlan"
--- a/usr/local/share/bastille/create.sh
+++ b/usr/local/share/bastille/create.sh
@@ -704,7 +704,7 @@ while [ $# -gt 0 ]; do
             ;;
         -v|--vlan)
 	    if echo "${2}" | grep -Eq '^[0-9]+$'; then
-                VLAN_ID="${2}
+                VLAN_ID="${2}"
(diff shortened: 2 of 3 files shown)


tschettervictor:
> Thanks @tschettervictor !
> 
> It seems I encountered multiple issues : 
> 
> 
> 
> `Template name/URL not recognized`
> 
> 
> 
> And my jail got the vnet0 with the IP but no VLAN (no vnet0.40)
> 
> 
> 
> I used `bastille create -B -V --vlan 40 test4`
> 
> 
> 
> By the way I realized that I have to respect this order which is strange to me, I tried to do 
> 
> `bastille create -B -V test4 --vlan 40 `and didn't work

Option must always come right after the command.

Did you update the "bastille.conf" file to include the VLAN template variable that the very bottom?

2Belette:
:+1:  very nice !! and many thanks

I was not aware about the order, all good.

I was also not aware about the bastille.conf!

All working great that's very nice, do you also plan to add the same --vlan switch to network command ? If would be nice when adding multiple vlan interfaces

tschettervictor:
> :+1:  very nice !! and many thanks
> 
> 
> 
> I was not aware about the order, all good.
> 
> 
> 
> I was also not aware about the bastille.conf!
> 
> 
> 
> All working great that's very nice, do you also plan to add the same --vlan switch to network command ? If would be nice when adding multiple vlan interfaces

Yes once this gets merged.

2Belette:
One more small thing I realized : bastille list -a is not showing any address when there is no address on vnet0 but one on vnet0.XX I don't think it is a bug, but I don't know what would be the best logic to display IP addresses when multiple interfaces are present in the jail.. perhaps trying vnet0 first with the vlan variant then vnet1 and so on ? there will never be a good response depending of each architecture...

Issue #618, opened 2023-10-25
[ENHANCEMENT] Execute command from template on host


tobiastom:
**Is your feature request related to a problem? Please describe.**
If I apply a template which requires a local directory, I need to create that manually and then run the template. 

**Describe the solution you'd like**
I would like to see a template command (e.g. `HOSTCMD`) that executes commands on the host system:

```
HOSTCMD mkdir -p /data/somewhere
FSTAB /data/somewhere /root/somewhere nullfs rw 0 0
```

**Describe alternatives you've considered**
Using [rocinante](https://github.com/BastilleBSD/rocinante), but that would be another dependency and a separate template which introduces template dependencies.

**Additional context**
This has been discussed before https://github.com/BastilleBSD/bastille/issues/348 https://github.com/BastilleBSD/bastille/issues/365

louwe:
+1 for this. This is also similar to what I'm requesting in #651.

tschettervictor:
Added.
#900

tobiastom:
Just for the record, `HOSTCMD` is not available due to security reasons discussed in #900.

I agree with them.

louwe:
I added a reply on that thread https://github.com/BastilleBSD/bastille/pull/900#issuecomment-2762015127 as to an actual use case beyond creating folder.

I'd imagine plenty of people  would like to be able to run webservers on standard port in the jail without having to run it as root.

tschettervictor:
@louwe lets keep the discussion to #900 just for simplicity and that we can discuss in one place.

Issue #365, opened 2021-03-24
[ENHANCEMENT] template for Master Host


kkalev:
Hello and thank you for your excellent work on BastilleBSD.

The template feature is quite compact and feature full in order to facilitate the initial setup of a Container (jail). It would be truly helpful if the administrator was able to leverage the feature in order to setup the Jailhost Master as well. Otherwise, one must use one solution (Puppet/Ansible) for the jailhost and another for the jails themselves.

My suggestion is to add a MASTER target in the "bastille template" command that would run the template on the jail host itself.

As a result my ideal action plan (in a virtualized environment) would be as follows:

1. Burn an OVA file using FreeBSD Cooker: https://github.com/aduitsis/freebsd-cooker
2. Deploy a thin puppet manifest/ansible playbook for an initial jail setup of the bastille pkg etc
3. Bootstrap and run a 'Server' template on the jailhost
4. Bootstrap FreeBSD release for containers
5. Create Container
6. Bootstrap a Template stack for the relevant container, ie 'Server' -> APM (Apache-PHP-MariaDB) -> 'Web Hoster' (include ProFTPd etc).

If the 3rd step was not available in Bastille the administrator would have to implement it in Ansible/Puppet (and have to learn and use both such a tool as well as templates for the jails themselves).

kkalev:
Ansible/puppet playbooks are basically needed only as a first step during/after cooking the FreeBSD OVA image in order to add ssh keys/bastille pkg etc.

Beyond that my need (and the reason for the request) is that I would prefer to only setup/maintain Bastille templates (for both the Jailhost and Jails) instead of both Ansible playbooks and Bastille templates.

For instance, both the Jailhost and the jails would need a basic server setup template (which would function as a first step in a template stack for more advanced jail setups for specific services). It seems an overkill to maintain basically the same commands/files/templates in two different formats/forms.

Thanks for the link.

yaazkal:
I get your point, but anyway for a "master template" or an autoconfiguration command like `bastille install` for the host, ansible/puppet is not needed. I think that can be done easily but I don't think that an external dependency will have a home here at least at this stage of the project. At this time that ansible playbook is the closest to what you are looking for.

kkalev:
Maybe I did not state my requirements clearly enough. I do not want Bastille to run any ansible/puppet playbooks.

[...]

PR #900, opened 2025-03-03
template: Add HOSTCMD to hooks

tschettervictor:
This allows specifying 

HOSTCMD echo "I am the host"

from a template to run the command on the host

### DIFF
--- a/docs/chapters/template.rst
+++ b/docs/chapters/template.rst
@@ -33,6 +33,8 @@ Template Automation Hooks
 +-------------+---------------------+-----------------------------------------+
 | CP/OVERLAY  | path(s)             | etc root usr (one per line)             |
 +-------------+---------------------+-----------------------------------------+
+| HOSTCMD     | command             | pkg info                                |
++-------------+---------------------+-----------------------------------------+
 | INCLUDE     | template path/URL   | http?://TEMPLATE_URL or project/path    |
 +-------------+---------------------+-----------------------------------------+
 | LIMITS      | resource value      | memoryuse 1G                            |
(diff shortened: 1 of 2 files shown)


yaazkal:
@tschettervictor @bmac2 what was the use case and motivation for this?

Templates should not run commands in the host, only in jails. This opens the possibility for insecure templates.

tschettervictor:
#618

Really?

I was not aware that templates should only function inside jails.

Thanks for pointing that out.

If you think, I will revert the changes.

yaazkal:
That issue also mentions #348 and #365

Look that for the host part, Rocinante was actually the solution (follow #365).

I think the host part is out of scope for the Bastille project. Discussion about if that is true or not is ok. Anyway a feature like that, if it's ever going to be implemented should be explicitly allowed by the Bastille user using a knob in the configuration file like `bastille_insecure_templates=YES`.

At the moment, the only use case is for creating a directory, not sure if that worths to open a pandora box. I vote for keeping the restriction.

Any thoughts on this @bmac2 @cedwards @JRGTH ?

tschettervictor:
I'm seeing your point security wise.

If no one has anything against it, I'll revert the change.

Better to keep the restriction while we discuss.

tschettervictor:
@yaazkal could weigh in here.

The reason we didn't do this is because of security reasons. We wanted the template to only apply inside the jail.

yaazkal:
> Some things in jail setups require running a command on the host. For example, if you want to run a web server in the jail on the traditional port 80 without having to run as `root`, you would need to install `security/portacl-rc` and create the`www` user/group on the host.

Which web server? nginx can run as 'nobody', it's just configuration.

Running commands in the host is out the scope of Bastille, take a look at [Rocinante](https://github.com/BastilleBSD/rocinante)

louwe:
> Which web server? nginx can run as 'nobody', it's just configuration.

It has nothing to do with the user. Yes, it can run as any user, but not on port 80. Any port below 1024 requires root user unless you install portacl on the host.

> Running commands in the host is out the scope of Bastille, take a look at [Rocinante](https://github.com/BastilleBSD/rocinante)

I would argue it's not if it is to directly support the automation of a jail, which is kind of the point of bastille and templates in the first place.

PR #916, opened 2025-03-30
minor fix to allow for display of multiple IP addresses from a jail w…

Tim-aslat:
Small fix to allow for jails with multiple IP addresses/interfaces.  The output from 'bastille list -a' was showing a newline with just the secondary IP which was causing errors for the json output as well.

This patch just concatenates IP addresses into a comma separated value


tschettervictor:
Hmmm. The reason for the change was to show the IPs each on a separate line.

Perhaps you could change it so that "if JSON=1; then" do what you suggest here.

tschettervictor:
@bmac2 I implemented JSON output.

bastille list -a -j

bmac2:
@Tim-aslat if you can fix it to @tschettervictor 's suggestion so it only does it in json format we can get this tested and merged before the release in the next couple of weeks.  I would consider this a bug so eligible to be merged even though we froze the code right now.

tschettervictor:
@Tim-aslat this will only work for VNET jails. See #917 and test it. It works for all types of jails.

bmac2:
superceded by 917 which does all that this does.

PR #919, opened 2025-04-03
list: Do not list Path and Hostname + list Name

tschettervictor:
#914 

This will now list Name after JID, and will not list Path or Hostname.

I've left the relevant code included in case we want to list it in the future.

This is a WIP.
@bmac2 @yaazkal 

Does it make sense to have "bastille list" call "jls"? I would argue that "bastille list" should list all jails bastille style without requiring the "-a|--all" flag. If someone wants to list using jls they can just type "jls".

### DIFF
--- a/usr/local/share/bastille/list.sh
+++ b/usr/local/share/bastille/list.sh
@@ -255,9 +255,6 @@ while [ "$#" -gt 0 ]; do
 	-h|--help|help)
 	    usage
 	    ;;
-        -a|--all|all)
-	    shift
-            ;;
 	-j|--json)
             OPT_JSON=1
 	    shift
@@ -283,9 +280,9 @@ while [ "$#" -gt 0 ]; do
 done
 
 # List json format, otherwise list all jails
-if [ "${OPT_JSON}" -eq 1 ]; then
+if [ "${OPT_JSON}" -eq 1 ] && [ "$#" -eq 0 ]; then
     list_all | awk 'BEGIN {print "["} NR > 1 {print "  {\"JID\": \"" $1 "\", \"Name\": \"" $2 "\", \"Boot\": \"" $3 "\", \"Prio\": \"" $4 "\", \"State\": \"" $5 "\", \"IP_Address\": \"" $6 "\", \"Published_Ports\": \"" $7 "\", \"Release\": \"" $8 "\","} END {print "]"}' | sed 's/,$//'
-elif [ "${OPT_JSON}" -eq 0 ]; then
+elif [ "${OPT_JSON}" -eq 0 ] && [ "$#" -eq 0 ]; then
     list_all
 fi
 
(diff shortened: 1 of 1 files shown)


2Belette:
Tested working fine, much better like this thanks

bmac2:
I think bastille list just calling jls is a mistake.  You don't get ip addresses with jls in most scenarios.  I would rather have the bastille list  style on list itself.  We can make list all do the path, name and as it is now.  

@yaazkal thoughts?

tschettervictor:
Good point. Although I'd just not do path and hostname at all.

tschettervictor:
I've modified the list command so it no longer requires "all" to list everything in bastille format.

Users can use "jls" if they want to see what "bastille list" used to show. I think this is far better.

bmac2:
I was thinking on this one today as i was driving.  How about this scenario.  We leave bastille list all alone.  let it ride as is.
But make bastille list show the bastille format without the full path so we can have a cleaner output.  Is there any other piece of data to add in the space where path is now?

tschettervictor:
This PR currently does not show path or hostname. I vote to keep it this way, and instead (if needed) add "list path" or "list hostname" etc if we want.

tschettervictor:
Update:

`bastille list` will now show short form
`bastille list all` will still show original format

yaazkal:
What if we have a header and list there the common things like path (option 1). Or if they happen to have different Path, then list them as references (option 2). This could be header or footer. I guess for the first option a header will work, for the second option a footer will do the trick even for releases. That way we keep the output as tight as possible and scriptable.

Option 1:

```
Bastille version: 0.14.20250420
Jails Path: /usr/local/Bastille/jails
JID   Hostname      Status
1     jail1         Up
2     jail2         Up
3     jail3         Down
```


Option 2:

```
JID   Hostname      Status    Realease   Path
1     jail1         Up        [A]        [1]
2     jail2         Up        [A]        [1]
3     jail3         Down      [B]        [2]
###
Bastille version: 0.14.20250420
Path [1]: /usr/local/Bastille/jails
Path [2]: /other/path/jails
Release [A]: 14.2-RELEASE
Release [B]: 13.5-RELEASE
```

Let's think about it before coding anything

tschettervictor:
Not that we want to copy other managers, but Appjail does it like this. It shows whatever you input as options. So if you would do "bastille list name jid path" then it would list only those 3 values.

We could also pick whatever we think is necessary to show by default and use "bastille list" for it. Then use "bastille list all" to show everything else. But that's almost too much.

bmac2:
I like the current way it is now doing.  Leave list all as it is and just list does all the parmeters without the path on it.  Nice and clean and easy.  All the footer notes that becomes a mess to maintain.

bmac2:
@yaazkal look at the way this PR does it now.  Simple, easy and does not invalidate the bastille list all, which some people actually use the path to the jail.  But beefs up the bastille list command to have all the parameters OTHER than the path.  I think this is a good compromise that doesn't mess anyone's workflow up.  What do you think?

2Belette:
just to cross check with you all, the new list seems very slow on my system, is it the same for you all ? (but I like it much better :)

tschettervictor:
The slowness probably comes from grabbing some of the variables using "bastille config"

I'll see if we can't do that in a faster way.

tschettervictor:
> Thanks ! not sure I see a difference, I would need to add more jails to check

Ok. I see a big difference here so...

2Belette:
any idea more or less of X amount of time to display Y number of jails in your list ?
just to give me an idea to compare

tschettervictor:
I can load 15 jails 2-3 seconds I'm guessing.

It's hard to time, it just goes one after the other fairly quickly.

tschettervictor:
Run "time bastille list"

Mine gives `2.404u 2.040s 0:03.23 137.4%    112+160k 0+0io 0pf+0w` for 16 jails, and it loaded pretty fast.

tschettervictor:
@2Belette Please test #1033

Try not to blink though...

2Belette:
Too fast for my eyes 😇

2Belette:
I pulled master and got the list slow again, I don't understand why as this is supposed to be merged in master...

2Belette:
Forgive me, I forgot to do a make install after the git pull master!
All good

PR #924, opened 2025-04-06
Implement parallel mode

tschettervictor:
This PR allows to run commands on multiple jails in parallel. Simply use '-p 3' to set a process limit. If not set, bastille will operate normally. The switch must follow the `bastille` keyword.

`bastille -p 3 start all` will start all jails but run 3 processes at a time.

This is a very rough implementation, and could be more streamlined, but here it is to test. 

@yaazkal @bmac2 What do you think?


bmac2:
tested and works as advertised.  Merging.

PR #930, opened 2025-04-10
Don't delete jail content on error, cleanup

JRGTH:
Don't delete jail content on busy datasets by default and some cleanup.

This bugfix will deal with the common "_cannot unmount 'XYZ': pool or dataset is busy_" which is an old error/bug who commonly happens in older/slow hardware when user try to delete a busy jail right after stopping them, also this may be override by the force option if the user what to go that route, otherwise will have a partially deleted/broken jail and manual intervention is required.

**Before fix:**
```
Deleting Jail: jail1.
cannot unmount '/mnt/storage/bastille/jails/jail1/root': pool or dataset is busy
cannot unmount '/mnt/storage/bastille/jails/jail1': pool or dataset is busy
rm: /mnt/storage/bastille/jails/jail1/root: Device busy
rm: /mnt/storage/bastille/jails/jail1: Device busy
Note: jail console logs archived.
/var/log/bastille/jail1_console.log-2025-04-09
```

**After fix:**
```
Deleting Jail: jail2.
cannot unmount '/mnt/storage/bastille/jails/jail2/root': pool or dataset is busy
cannot unmount '/mnt/storage/bastille/jails/jail2': pool or dataset is busy
Jail [jail2] dataset(s) appears to be busy, exiting.
```

### DIFF
--- a/usr/local/share/bastille/destroy.sh
+++ b/usr/local/share/bastille/destroy.sh
@@ -48,16 +48,16 @@ EOF
 
 destroy_jail() {
 
-    local OPTIONS  
+    local OPTIONS
 
     for _jail in ${JAILS}; do
-	
+
         bastille_jail_base="${bastille_jailsdir}/${_jail}"            ## dir
         bastille_jail_log="${bastille_logsdir}/${_jail}_console.log"  ## file
-	
+
         check_target_is_stopped "${_jail}" || if [ "${AUTO}" -eq 1 ]; then
             bastille stop "${_jail}"
-        else   
+        else
             error_notify "Jail is running."
             error_continue "Use [-a|--auto] to auto-stop the jail."
         fi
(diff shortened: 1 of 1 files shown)


bmac2:
What is the best way to test this.  I am trying to create the error before testing your code and can't seem to create the error.  Any suggestions on how to recreate the error?

JRGTH:
> What is the best way to test this. I am trying to create the error before testing your code and can't seem to create the error. Any suggestions on how to recreate the error?

Hi Barry nice to hear you bro \m/, this error is a pretty old bug and it was workaround with the `-f`option but it should not happen on the first lace regardless of the options used.

An example on how to trigger this issue is to creating a "Thick VNET" jail and right after creation execute  `pkg update` with the command `bastille cmd jail1 pkg update` and right after finish update stop the jail and quickly try destroy the jail, so some datasets/directories of the jail1 may still be open/used by the host hence triggering this error early, notoriously on slow devices.

This error does happens to me alot and I meant often when I'm using my scripts for quick testing stuff and had to use `-f` flag for quick jail deletion which is not always desirable, otherwise I have to wait few seconds to a minute to delete the jail normally, if I delete with the `-f` option the jail will be partially deleted/broken and I have to manually unmout the datasets and destroy them manually, haven't checked deeper on what shell are temporarily holding the directories though.

Regards!

tschettervictor:
> > What is the best way to test this. I am trying to create the error before testing your code and can't seem to create the error. Any suggestions on how to recreate the error?
> 
> 
> 
> Hi Barry nice to hear you bro \m/, this error is a pretty old bug and it was workaround with the `-f`option but it should not happen on the first lace regardless of the options used.
> 
> 
> 
> An example on how to trigger this issue is to creating a "Thick VNET" jail and right after creation execute  `pkg update` with the command `bastille cmd jail1 pkg update` and right after finish update stop the jail and quickly try destroy the jail, so some datasets/directories of the jail1 may still be open/used by the host hence triggering this error early, notoriously on slow devices.
> 
> 
> 
> This error does happens to me alot and I meant often when I'm using my scripts for quick testing stuff and had to use `-f` flag for quick jail deletion which is not always desirable, otherwise I have to wait few seconds to a minute to delete the jail normally, if I delete with the `-f` option the jail will be partially deleted/broken and I have to manually unmout the datasets and destroy them manually, haven't checked deeper on what shell are temporarily holding the directories though.
> 
> 
> 
> Regards!

Happens to me a lot as well. Thanks for this @JRGTH

PR #940, opened 2025-04-12
bastille setup command rewrite/revisited

JRGTH:
This setup helper command may be of benefit for newcomers to setup bastille with some sane/predefined parameters.

This is PR has been re-open as requested.

### DIFF
--- a/usr/local/share/bastille/setup.sh
+++ b/usr/local/share/bastille/setup.sh
@@ -105,7 +105,7 @@ config_backup() {
 config_network_reset() {
     # Restore bastille default network options.
     warn "Performing Network configuration reset, requested by the user..."
-    read -p "Do you really want to reset 'bastille' network configuration? [Y|n]: " _response
+    read -rp "Do you really want to reset 'bastille' network configuration? [Y|n]: " _response
     case "${_response}" in
         [Yy]|[Yy][Ee][Ss])
             config_backup
@@ -129,7 +129,7 @@ config_network_reset() {
 config_storage_reset() {
     # Restore bastille default ZFS storage options.
     warn "Performing ZFS configuration reset, requested by the user..."
-    read -p "Do you really want to reset 'bastille' ZFS configuration? [Y|n]: " _response
+    read -rp "Do you really want to reset 'bastille' ZFS configuration? [Y|n]: " _response
     case "${_response}" in
         [Yy]|[Yy][Ee][Ss])
             config_backup
@@ -159,7 +159,7 @@ config_restore_global() {
     # Be aware that if the sample configuration file is missing, we can generate a new one,
     # but that's highly unlikely to happen so will keep the code smaller here.
     warn "Performing Bastille default configuration restore, requested by the user..."
-    read -p "Do you really want to restore 'bastille' default configuration file and start over? [Y|n]: " _response
+    read -rp "Do you really want to restore 'bastille' default configuration file and start over? [Y|n]: " _response
     case "${_response}" in
         [Yy]|[Yy][Ee][Ss])
             config_backup
@@ -347,7 +347,7 @@ zfs_initial_activation() {
     # Just let the user interactively select the ZFS items manually from a list for the initial activation.
     # This should be performed before `bastille bootstrap` as we already know.
     info "Initial bastille ZFS activation helper invoked."
-    read -p "Would you like to configure the bastille ZFS options interactively? [Y|n]: " _response
+    read -rp "Would you like to configure the bastille ZFS options interactively? [Y|n]: " _response
     case "${_response}" in
         [Yy]|[Yy][Ee][Ss])
             # Assume the user knows what hes/she doing and want to configure ZFS parameters interactively.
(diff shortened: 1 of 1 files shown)


JRGTH:
Re-opened the rewrite of the `bastille setup` command helper as requested.

**Overview:**
```
root@freebsd14:~ # bastille setup 
Usage: bastille setup [option]
    Options:

    -p | --firewall                    Attempt to configure bastille PF firewall.
    -l | --loopback                    Attempt to configure network loopback interface.
    -e | --ethernet                    Attempt to configure the network shared interface.
    -v | --vnet                        Attempt to configure VNET bridge interface [bastille1].
    -z | --zfs                         Activates ZFS storage features and benefits for bastille.
         --zfs-custom-setup            Manually configure ZFS, this is intended for advanced users.
         --conf-network-reset          Restore bastille default Network options on the config file.
         --conf-storage-reset          Restore bastille default ZFS storage options on the config file.
         --conf-restore-clean          Restore bastille default config file from bastille.conf.sample file.
```

Options are self explanatory and the ZFS configuration is straight forward for installations through FreeBSD pkg tool, with exception for the `--zfs-custom-setup` which is intended for advanced users.

However manual ZFS setup options may also be triggered under UFS systems as expected as the user has to point for the correct ZFS pool/parameters as expected.

Also each config change will create a backup file as: `bastille.conf.YYYYMMDD-HHMMSS` for convenience.

Regards

bmac2:
@JRGTH  there are now conflicts that need to be resolved.

bmac2:
first comment.  running the setup the first time should act as it does now,  IF it detects you are running zfs it sets it up for  you.  That HAS to remain.  Having all the other options is fine, but if a newbie runs setup the first time they don't need to have to run your zfs helper to get zfs working when we have simple code to do that out of the box.

bmac2:
second comment:  setup as it is now creates the bastille0 loopback by default.  Why should I have to run setup multiple times just to get zfs and loopback there??  This is overly complicated for a new user.

bmac2:
third comment:  What you have as -a or auto should be the default that runs when you run bastille setup.  Then you can run the others if you are tweaking settings.  For a newbie we need to keep it so they run bastille setup and they have a working system with bastille0 and zfs if they are running zfs.  That has to remain a minimum.  

@yaazkal @tschettervictor @cedwards

bmac2:
running the vnet command to create the bridge for vnet bridged containers failed and did not work on my system.

bmac2:
what is -e trying to do?  it picked out my vtnet0 but when it ran it made no changes.  What is the "network shared interface" and what does that mean?

JRGTH:
> third comment: What you have as -a or auto should be the default that runs when you run bastille setup. Then you can run the others if you are tweaking settings. For a newbie we need to keep it so they run bastille setup and they have a working system with bastille0 and zfs if they are running zfs. That has to remain a minimum.
> 
> @yaazkal @tschettervictor @cedwards

Hi Barry nice to hear you man, yes the `bastille setup -a` can be made to act like the previous `setup.sh` indeed, as for the "vnet loopback" not working,  the command does apply the exact same bits the old `setup.sh` command already does so I'm not aware of any issue.

As for the ZFS helper this bits are for the sole user who are not running a 100% ZFS on they host, many users still use UFS for the operating system, hence this users can manually use the ZFS helper for setup bastille storage target zpoo/dataset. 

As for the `${bastile_network_shared}` this is for the users who do not run VNET, loopback nor PF/firewall etc. and they just want a shared network interface to simply run "Shared IP" jails, jails who operated with a shared static IP and does not feature complex networking.

I can make all this fixes in a bit but as I previously denoted, apart from the ZFS setup helper, every other bits are an exact inclusion of the previous setup.sh command, I've not altered vnet, loopback, bridge as I don't use those on most of my simple jails.

PS. Since my bastille ZFS setup is different from the average user who just run `pkg install bastille` in a basic ZFS system, I have dedicated ZFS Mirrors, RAIDZ2 with rolling snapshots etc, and If I execute the `setup.sh` command my `bastille.conf` just gets borked, so this setup command needed this rework, I've previously reported such issue but can't find the comments, between I will revise again the VNET, Loopback, Bridge bits and compare against the previous `setup.sh` and make sure all is exactly the same.

Regards

JRGTH:
Ahh yeah, looks like the previous `setup.sh` command has been updated after I've made some rewrite indeed, hence it does have the old code triggering some conflicts with current code, well I've also vote to keep this command straight simple and lean as well.

So I've concluded that the ZFS helper inclusion may be of very little to no benefit for users running UFS on the OS and ZFS for storage, since most of this users are seasoned FreeBSD already, so lets keep the previous and simple `setup.sh` since this rewrite will nearly add 50kB of code and provably of very little use other than bloat, not to mention that newcomers will have very little to no benefit from the options menu as this command mean to be run only once after fresh `bastille` installation with sane defaults.

Between thanks for the suggestions and for spotting this network issues in this rewrite, unfortunately I'm a bit busy ATM so can't deeply test more in the networking/firewall area.

Regards

JRGTH:
I vote to keep this initial setup command simple and straight forward for newcomers.

tschettervictor:
> I vote to keep this initial setup command simple and straight forward for newcomers.

I'm actually in favour of this PR. It brings an interactive experience to the end user.

All that's needed is to have "bastille setup" just run all the functions as it did previously.

tschettervictor:
I think we are confusing something. Inside the config file there are two variables that can be used to configure the default interface.

bastille_network_loopback
bastille_network_shared

I vote to merge these two into the following.

bastille_network_default_interface

The reason is that there is really no point in having both values.

The create command (if no interface is supplied) expects only one of these two values to be populated. 

My argument is, what's the point of having both? 

Each of these two are only ever used if you don't supply an interface in your create command. So let's make it "bastille_network_default" and let it do what it's supposed to. And that is to be the default interface if none is supplied.

PR #951, opened 2025-04-16
Fix epair numbering when mixing -V and -B jails on one systme

tschettervictor:
This fixes a bug when using VNET and bridged VNET jails on the same system.


bmac2:
tested pretty extensively.  Fixes issue with mixing B and V jails on one system.  Epairs now count correctly and don't try to use the same epairs.

PR #962, opened 2025-04-22
temporary setup command fix

tschettervictor:
This is simply to avoid issues when the setup command might be run over an existing setup.

It will check each setup subcommand and verify it has not already been configured before attempting to run the setup for the relevant setup step.

This is only temporary, as we would like to get a much more advanced setup command integrated in this or below.

#940


bmac2:
tested and works and is an improvement to the current setup command.

@JRGTH  test this and see if it meets your requirements

@yaazkal @tschettervictor

Issue #262, opened 2020-10-08
[ENHANCEMENT] Jail VNET with Netgraph


PetruGarstea:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
N/A

**Describe alternatives you've considered**
Any plans to implement **VNET Jails** with **Netgraph** ?

ref: 

- /usr/share/examples/jails/jng
- https://wiki.freebsd.org/TomMarcoen/JailNetworking

**Additional context**
In my setup the current traffic accounting is done with [ng_netflow](https://www.freebsd.org/cgi/man.cgi?query=ng_netflow&apropos=0&sektion=4&manpath=FreeBSD+12.1-RELEASE&arch=default&format=html), traffic passed from **epair** interfaces via **if_bridge** is not being counted, I am wondering if change the networking to **jng** would **ng_netflow** count the traffic originated from VNET jails.

PetruGarstea:
Greetings,

I spent last week on exploring **netgraph** networking for jails and here what I want to share/propose for this **ENHANCEMENT**

1. Setting up **netgraph** or **if_bridge** networking should come from the main config file in the similar way how the user chooses the file system, between **ZFS** or **UFS** 

2. In my POC tests **ng_bridge** and **if_bridge** should not co-exits on the same host, that means user has only one option to choose and for the migrated jail it needs to be detected what type of networking on the host is running.

3. Netgraph ethernet interfaces for jails can be managed by **jng** script that is located under **/usr/share/examples/jails/jng**

4. Netgraph modules have to be loaded into kernel, 

```
cat /boot/loader.conf

netgraph_load="YES"
ng_netflow_load="YES"
ng_ksocket_load="YES"
ng_ether_load="YES"
ng_bridge_load="YES"
ng_eiface_load="YES"
ng_socket_load="YES"
```

5. netgraph config, that creates bridge and exports netflow that are passes via main physical ethernet interface

```
# send from physical interface to netflow node
ngctl mkpeer lan: netflow lower iface0
# name netflow node as netflow
ngctl name lan:lower netflow
# return from netflow to physical interface
ngctl connect lan: netflow: upper out1
# connect netflow node with bridge node
ngctl mkpeer netflow: bridge out0 link0
# name bridge node as lanbridge
ngctl name netflow:out0 lanbridge
# connect bridge node with netflow node
ngctl connect lanbridge: netflow: link1 iface1
# create management interface for tests 
[...]

PR #978, opened 2025-04-25
bastille: Initial support for netgraph

tschettervictor:
#262 

This introduces initial support for netgraph using the JNG script included in a default BSD install. 

This requires a host that is brand new. Netgraph DOES NOT work with epairs. You must use one or the other. 

To get started.

1. Run `bastille setup netgraph` to load the required modules. This will also set the new variable inside bastille.conf `bastille_network_vnet_type=netgraph`. Don't forget to diff update your old config file.

2. Create a jail using -V. -B is not supported. 

EDIT: Create, rename, clone, network, import subcommands should all work with netgraph at this point. Bug reports are welcome.

This is highly experimental, but it seems to work fine.


tschettervictor:
@adriel-tech 
@patmaddox 

Can you guys test and see if this works?

bmac2:
tested and working  merging

PR #984, opened 2025-04-28
list: First draft of using config to get variables

tschettervictor:


bmac2:
tested and it looks beautiful!!!

PR #1003, opened 2025-05-02
Initial beta support for dependant jails

tschettervictor:
This will add support for `bastille config TARGET set depend jail1`

When starting TARGET, jail1 should be started before TARGET starts.

When stopping jail1, TARGET should be stopped before jail1 stops.

This is simply a value at `${bastille_jailsdir}/${TARGET}/settings.conf` which can also be edited manually.

This is beta, so bug reports are welcome.

The `depends` value will be auto filled with and empty value when jails are first stopped, or first created.

### DIFF
--- a/usr/local/share/bastille/config.sh
+++ b/usr/local/share/bastille/config.sh
@@ -157,14 +157,14 @@ for _jail in ${JAILS}; do
     # Depend property
     elif [ "${PROPERTY}" = "depend" ] || [ "${PROPERTY}" = "depends" ]; then
 
-        PROPERTY="depends"
+        PROPERTY="depend"
         BASTILLE_PROPERTY=1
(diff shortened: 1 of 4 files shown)


bmac2:
tested extensively and works as advertised.  Sets up dependencies for jails and on start, restart and stop it works as advertised.

2Belette:
That's a nice addition!

@bmac2 I haven't been able to fest it yet but I am wondering if multiple dependencies had been tested ?
Like jail1 (web) depends to jail2 (db) which also depends on jail3 (syalog server) ?

tschettervictor:
> That's a nice addition!
> 
> 
> 
> @bmac2 I haven't been able to fest it yet but I am wondering if multiple dependencies had been tested ?
> 
> Like jail1 (web) depends to jail2 (db) which also depends on jail3 (syalog server) ?
> 
> 

This should work.

PR #1018, opened 2025-05-04
Parallel subshell

tschettervictor:
This changes the parallel mode to simply use subshells and a function to track running jobs.
Cleaner and more efficient than xargs.


bmac2:
tested and working

PR #1019, opened 2025-05-05
Implement jail migration to remote system

tschettervictor:
`bastille migrate TARGET USER HOST`

Should export a jail, send to a remote host, then import. -a to have it start on the remote side. -d to destroy the old jail. 

To test, you must 

Pull the PR

Diff the config file (new 'bastille_migratedir' variable) on both systems

Have a user that can access your remote system with a key file (no password support just yet)


bmac2:
tested and working.  STILL experimental and beta status

PR #1021, opened 2025-05-05
pkg: Fix promt for y|n

tschettervictor:


bmac2:
tested and bug with parallel processing fixed.

PR #1023, opened 2025-05-06
Fix `route` command for IPv6

gahr:
Route needs to know we're passing it an IPv6 address:

```
# route -n get fec0:0:0:3::80
route: bad address: fec0:0:0:3::80
```


tschettervictor:
This breaks the IPv4 part though.

gahr:
How does it break the IPv4 part? The IPv4 logic is almost identical but it's the block of code above.
https://github.com/BastilleBSD/bastille/blob/5db3ec1764ba4a21dd95cd1c1aee82859e853825/usr/local/share/bastille/start.sh#L138-L160

tschettervictor:
My bad. I'm still sleeping...

PR #1024, opened 2025-05-06
Fix `console` with a user

gahr:
Sorry, boys, but do you even *minimally* test your changes?


tschettervictor:
Good point. I personally never use the "user" option for the console, but thanks for the call out and reminder.

gahr:
Thanks for taking my feedback positively. I realize I could have toned it down a bit.

PR #1032, opened 2025-05-08
migrate: Use ${USER} as user

tschettervictor:


2Belette:
I am going much further with this one but still see some issues : 

+ ssh -p 22 -i /home/blt/.ssh/id_rsa blt@10.7.30.1 bastille list jails
Bastille: Permission Denied
root / sudo / doas required
+ local '_remote_jail_list='
+ echo ''
+ grep -Eoqw jail-c3
+ checkyesno bastille_zfs_enable
+ eval '_value=$bastille_zfs_enable'
+ _value=YES
+ return 0
+ checkyesno _remote_bastille_zfs_enable
+ eval '_value=$_remote_bastille_zfs_enable'
+ _value=YES
+ return 0

should not be + ssh -p 22 -i /home/blt/.ssh/id_rsa blt@10.7.30.1 doas bastille list jails instead ?

at the end it fails because of 
```

+ scp -P 22 -i /home/blt/.ssh/id_rsa /usr/local/bastille/migrate/jail-c3_2025-05-08-150616.sha256 blt@10.7.30.1:/usr/local/bastille/migrate
scp: dest open "/usr/local/bastille/migrate": Permission denied
scp: failed to upload file /usr/local/bastille/migrate/jail-c3_2025-05-08-150616.sha256 to /usr/local/bastille/migrate
```

is there a need to have the user owns the directory ?

tschettervictor:
I think we need to adjust the permissions on the migrate dir.

And I think we need to add the super user to the list jails cmd.

tschettervictor:
@2Belette Guve it another go.

Now it will check and make sure perms are 777.

Added the -w 1 also.

2Belette:
I see that 

`drwxrwxrwx   2 root wheel  2 May  7 22:36 migrate`

but

```
+ scp -P 22 -i /home/blt/.ssh/id_rsa /usr/local/bastille/migrate/jail-c3_2025-05-08-212446.sha256 blt@10.7.30.1:/usr/local/bastille/migrate
scp: dest open "/usr/local/bastille/migrate": Permission denied
scp: failed to upload file /usr/local/bastille/migrate/jail-c3_2025-05-08-212446.sha256 to /usr/local/bastille/migrate
```

tschettervictor:
> I see that
> 
> `drwxrwxrwx 2 root wheel 2 May 7 22:36 migrate`
> 
> but
> 
> ```
> + scp -P 22 -i /home/blt/.ssh/id_rsa /usr/local/bastille/migrate/jail-c3_2025-05-08-212446.sha256 blt@10.7.30.1:/usr/local/bastille/migrate
> scp: dest open "/usr/local/bastille/migrate": Permission denied
> scp: failed to upload file /usr/local/bastille/migrate/jail-c3_2025-05-08-212446.sha256 to /usr/local/bastille/migrate
> ```

The directory should be wide open for any user. It is just used as a temporary storage for the jail archives. I'm not sure why it is giving you the permission denied there.

Can you try deleting the migrate directory and letting bastille recreate if for you? Or dataset if using ZFS.

2Belette:
after `doas zfs destroy zroot/bastille/migrate` I got the same issue/error message (and I don't see Bastille to recreate the dataset)

tschettervictor:
Run bastille setup -f then try again.

Did I miss that in migrate?

tschettervictor:
@2Belette Did you check the local and remote sides for perms?

tschettervictor:
I just tried the same setup, and it was able to complete the migration. I purposely changed the perms to 644 on the migrate directory, and it detected that and changed it back to 777.

2Belette:
I made more tests and, the good : if I destroy zfs dataset migrate and the folder, they are recreated with the 777 permissions.

So I tried to touch a file in /usr/local/bastille/migrate with my user and got permission denied, strange for a 777 isn't it ..?

And I took a look on /usr/local/bastille permission and it was drwxr-x---   8 root wheel so no folder traversal permission, I added doas chmod o+x /usr/local/bastille and got the migrate successful !

tschettervictor:
I have an idea

Let's use an /tmp/bastille-migrate for the migration with an option to retain the archives on the remote system.

This would copy the archives to the backups directory of Bastille, then rm the tmp directory we use for migration.

This will also eliminate the need for a migrate directory variable.

tschettervictor:
@2Belette 

Test again, with a fresh install. We do not need the bastille_migratedir anymore. This way we don't have to introduce breaking changes.

Ive also add --backup if you want to keep the archives on the remote system. They will get copied into the backups dir.

2Belette:
Thanks will do the test tomorrow. That's fun, I was thinking about backup, I never used the function, but I am wondering if the exact same code could be use to do remote backup? The only thing would be to stop the jaik, do the migrate function, and re start the jail (and not starting the jail on the remote server). Just an idea

tschettervictor:
> Thanks will do the test tomorrow. That's fun, I was thinking about backup, I never used the function, but I am wondering if the exact same code could be use to do remote backup? The only thing would be to stop the jaik, do the migrate function, and re start the jail (and not starting the jail on the remote server). Just an idea

That would have to be a job for an API.

You can simply run "bastille export TARGET" on any host to create a backup of any jail.

tschettervictor:
By default, migrate will only export, send, and import the jail. It will no start. If you want it to start, you can enable `-a|--auto` for auto mode.

2Belette:
Yes, I am always using -a which stops the jail and starts in the remote.
The trick would be to stop first, doing the migrate without -a and start the jail again.
I am wondering, is there any attempts in the past to try to do "hot migrate/backup" ? I imagine data integrity is the potential issue, but with zfs it is perhaps safer, or is it a because of memory ? This would be a nice feature to be closest to HA

tschettervictor:
Was think about adding the `--live` option as we do in the clone command.

I think it's a neat idea.

tschettervictor:
Implemented.

`-l|--live` will now allow migrating a running ZFS jail. 
If you have `--auto` set, then the jail will start on the remote side.

`--live` takes precedence over `--auto` so if both options are set, the migration will happen "live" or "hot"

2Belette:
That's very nice ! I tested it and it works perfectly.

I am just questioning the behavior : 

without --auto, it does the live migration and the migrated jail on the remote is down, it is OK.

The question is when we combine --auto and --live, it keeps the jail ON and also put the jail ON on the remote host... I am wondering if we want this to happen ? I understand perfectly the logic in term of bastille commands but in term of network point of view... both will have the same network parameters (same IP, etc...) into two different hosts, which is not something we should ever do.

perhaps --auto and --live should not be used at the same time.
--live should do what we do now
and perhaps a --live --off should migrate, turn the current OFF and put the migrated on the remote system ON, like a hot migration

tschettervictor:
What about this?

Live only = local on, remote off
Auto only = local off, remote on
Live+auto = (migrate while on) then local off, remote on
None = both off

tschettervictor:
We are settling for the following.

auto will always start the remote jail. If paired with live, the old jail will be stopped after the migration

live on its own will leave jails as they are, running locally and stopped remotely.

2Belette:
that's very cool many thanks !!! Tested and working perfectly. I like the way it interacts between auto and live

PR #1035, opened 2025-05-09
Support jailing datasets

tschettervictor:
This is an initial testing for jailing ZFS datasets.

First, create a dataset you want to jail. You don't have to set the `jailed=on` property. Bastille does that.

Then run `bastille zfs TARGET jail pool/dataset /path/in/jail`. This will set the config variables, and add the entry to `zfs.conf` at the jail root. Once the jail starts, these entries will be read and added to the jail.

When you start the jail, bastille will `jail`, and `zfs mount` the specified dataset at the specified mount point inside the jail.

When the jail is stopped, bastille will `zfs umount`, and `unjail` the dataset (in that order), and then the jail will be stopped.

This is BETA, 
Do not attempt to add and remove additional datasets to the `zfs.conf` file while the jail is running, unless you know what you are doing. Stopping a jail without unmounting the datasets will result in permission errors the next time the jail is started.

To remove a dataset, run `bastille zfs TARGET unjail pool/dataset`

### DIFF
--- a/docs/chapters/zfs-support.rst
+++ b/docs/chapters/zfs-support.rst
@@ -82,3 +82,34 @@ the file is simple.
   pool/other/dataset /other/path/in/jail
 
 To remove a dataset from being jailed, we can run ``bastille zfs TARGET unjail pool/dataset``.
+
+Template Approach
+^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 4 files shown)


tschettervictor:
@dch 
@g-teley
@patmaddox 
@markjdb

PR #1067, opened 2025-05-16
setup: Fix UFS setup

tschettervictor:


bmac2:
tested and verified it now sets up ufs correctly even  on a ufs vm on a zfs host.

Issue #1084, opened 2025-05-20
[ENHANCEMENT] set defaultrouter and nameserver on jail creation


thenktor:
**Is your feature request related to a problem? Please describe.**
When setting up jails for various VLANs I have to change settings for defaultrouter and nameservers afterwards because they obviously differ from the host's settings.

**Describe the solution you'd like**
I'd like to specify defaultrouter and nameserver via command line arguments on jail creation, e.g. `bastille create --vnet --defaultrouter 192.168.123.1 --nameserver 192.168.123.2 caddy 14.2-RELEASE 192.168.123.10/24 vlan123`

tschettervictor:
This should be working with #1087 

Can you confirm?

@thenktor

thenktor:
Hi,

is it possible to run bastille from the downloaded git directory without installing it?
It seems all pathes to the source files are hardcoded to absolute pathes, e.g. `. /usr/local/share/bastille/common.sh`.

2Belette:
I second that !

PR #1087, opened 2025-05-20
create: Add -g|--gateway

tschettervictor:


2Belette:
Don't know if it is by designed but I got two times the same entry in my resolv.conf.
Is there a way to have the option to support or one entry or two entries ? This should cover most of the usecases

tschettervictor:
Hmmm. You should not have two entries. 

Can you share the resolv.conf file?

2Belette:
See the comment here : https://github.com/BastilleBSD/bastille/issues/1084#issuecomment-2898120525

thenktor:
The patch replaces all occurences of nameserver with the provided one.

`--nameserver 9.9.9.9` on this resolv.conf:
```txt
nameserver 1.1.1.1
nameserver 2.2.2.2
```

results in:
```txt
nameserver 9.9.9.9
nameserver 9.9.9.9
```

Actually you should remove all lines starting with nameserver: `/^nameserver.*/d`
Then add the one.

Better: Make it possible to add multiple servers with a comma separated list.

tschettervictor:
Done.

I was not aware that you had multiple nameservers.

You should be able to now specify a comma separated lists also.

2Belette:
Maybe I am not using the option correctly but `--nameserver 1.1.1.1,8.8.8.8` is producing a /etc/resolv.conf with :

 `nameserver 1.1.1.1,8.8.8.8`

this should be : 
```
nameserver 1.1.1.1
nameserver 8.8.8.8
```

tschettervictor:
My bad. Fixed.

PR #1093, opened 2025-05-21
template: Pass ARG_FILE to INCLUDE templates

tschettervictor:
@tobiastom 

This should now pass `--arg-file` to any INCLUDE templates. 

Unfortunately we might not be able to support using `--arg ARG=VALUE` because there are times when we don't want to pass the same ARGS to the INCLUDE template. For example, we are writing some unit tests that are 3 layers deep, and call different ARGS at different times.

One option would be to include a `--recursive` options in the template command. Do you think that would work ok, or should we just pass the arg file by default?

### DIFF
--- a/docs/chapters/subcommands/template.rst
+++ b/docs/chapters/subcommands/template.rst
@@ -21,5 +21,6 @@ The TEMPLATE arg should be called with the ``project/template`` format.
   
       Options:
 
-      -a | --auto           Auto mode. Start/stop jail(s) if required.
-      -x | --debug          Enable debug mode.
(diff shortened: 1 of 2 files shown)


tobiastom:
@tschettervictor I think I'm a little lost. You did stop implementing this in the end, right? 

I think I'd like the `--recursive` argument, but only for `--arg-file`. That said I'm not sure if that's all that intentional, and I think I'd rather stick to passing --arg myself to make it more explicit.

Also, the `--arg` already works, right? This is why you closed it? 😆

tschettervictor:
Correct on all counts.

PR #1130, opened 2025-06-09
Dynamically create epair in bridge mode

nostrad:
Previously bastille in existing bridge mode created a static index for the epair. This has the unfortunate side effect of not handling existing epairs or new epairs being created outside bastille.

Instead, use the returned name and create a bastille unique interface name to use rather than epair<NUM>.

### DIFF
--- a/usr/local/share/bastille/common.sh
+++ b/usr/local/share/bastille/common.sh
@@ -142,7 +142,7 @@ get_bastille_if_count() {
     if [ "${bastille_network_vnet_type}" = "if_bridge" ]; then
         for _config in /usr/local/etc/bastille/*.conf; do
             local bastille_jailsdir="$(sysrc -f "${_config}" -n bastille_jailsdir)"
-            _bastille_if_list="$(printf '%s\n%s' "$( (grep -Ehos '(epair[0-9]+|bastille[0-9]+)' ${bastille_jailsdir}/*/jail.conf; ifconfig -g epair | grep -Eos "_bastille[0-9]+$"; ifconfig -g epair | grep -vs 'bastille' | grep -Eos 'e[0-9]+a_') | grep -Eos '[0-9]+')" "${_bastille_if_list}")"
+            _bastille_if_list="$(printf '%s\n%s' "$( (grep -Ehos '(epair[0-9]+|bastille[0-9]+|e[0-9]+a_)' ${bastille_jailsdir}/*/jail.conf; ifconfig -g epair | grep -Eos "_bastille[0-9]+$") | grep -Eos '[0-9]+')" "${_bastille_if_list}")"
         done
         _bastille_if_count=$(printf '%s' "${_bastille_if_list}" | sort -u | wc -l | awk '{print $1}')
         export _bastille_if_list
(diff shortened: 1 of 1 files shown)


nostrad:
This "works for me" and allows existing epairs to coexist fully with the epairs bastille wants to create, getting me out of the "epair0 already exists" error (since my epair happens to not be named e#a_ in ifconfig).

The `get_bastille_if_count` function is a bit unclear to me what all the greps attempt to achieve but I think this could solve the generic case.

tschettervictor:
The current code does what you want already. 

It count all epairs (created by Bastille and not created by Bastille) and increments by one for any new jail.

We have had this issue in the past, but as of the latest code, it should be resolved.

Are you testing with the latest code?

tschettervictor:
But as far as your epairs not being named with a number, I'm not sure if this is the best approach here.

We will have to discuss.

tschettervictor:
So, I don't believe you are solving anything. Reasons below...

1. We are already (latest code) counting the epairs that are in use in your system with the "get_bastille_if_count". We do this by count an "eXa" in jail.conf, and using "ifconfig -g epair" to count any that may be in use outside of Bastille.

2. This will still leave the numbers static as you are not touching the "$host_epair" and "$jail_epair" variables. These will still be static. As I see it it could actually cause issues if you do "ifconfig epair create" and it creates "epair6" but the value is already hardcoded as "epair10" or whatever.

Am I understanding that your ultimate issue is that you have epairs in use on your system that Bastille is not managing?

If so, then we are already counting those in the "get_bastille_if_count". If this meant working then there is a bug.

I think it best to stay as it is for the sake of keeping the jail.conf file simple and easy to follow.

nostrad:
But you grep out eXa and epairX names from that list, here's an example that will fail

```
ifconfig epair0 create
ifconfig epair0a name ep_jail0
ifconfig epair0b name ep_jail0_vnet
```

Now you can't use bastille, because it tries to create epair0, and that name is now reserved!

AND, if you create epairs externally, the bastille code has statically allocated epair0 for the jail and the jail cannot be started anymore.

nostrad:
What I am trying to do here is to make bastille only put its own indexing on interfaces and as long as the user does not manually create eN[ab]_[jailname] interfaces it won't clash (the user can still create e0a_foo, as long as foo is not identical to the bastille jail name)

tschettervictor:
I'm not sure I follow. Bastille always keeps the number part intact of the epairs. So all epairs that are managed by Bastille will include that number.

With the grep commands we are gathering all of those numbers, then creating an epair with the first number not present sequentially.

I get that it will fail if there are no numbers present in the name, but there really isn't a way around that without some more discussion.

May I ask why you aren't keeping the number of the epair as part of your manual epair names?

nostrad:
because they are named as ep_foo_0_br, and ep_foo_0_vnet, because that made sense when I set things up, and no grep in the world can actually be future proof as my _foo and _bar shows...

By doing `epair_name=$(ifconfig epair create)` we do not need to look at ifconfig at all for numbering, the only thing that matters is that bastille internally has unique names when it creates its renamed interfaces.

And it will also be safe against some other tool creating an epair, such as jib which I believe creates epairs with dynamic numbers! So it might very well be that bastille can interfere with itself...

EDIT, now I'm curious, I wonder what happens if you do
```
bastille create -B testA 14.2-RELEASE 192.168.0.10/24 bridge0
bastille stop testA
bastille create -V testB 14.2-RELEASE 192.168.1.10/24 vtnet0
```
I think testB will have jib create epair0 in that case and now you won't be able to start testA... Need to setup a VM to test that one...

nostrad:
But it will not work, the problem is the epair before the rename, if jib creates an epair from the first free name, and renames it to e1a_something, internally it will still be epair0, thus ifconfig epair0 create will fail.

Creating them dynamically is what solves this, the "epairN" name is dynamic and then it gets renamed to the statically allocated name, which is much less conflict prone

tschettervictor:
I understand.

But, epairs always get named "epairN" to begin with, so the actual number is still in record in the "dmesg". Even after a rename it is still there.

```
oot@dev1:~ # ifconfig epair create
epair7a
root@dev1:~ # ifconfig epair7a name ea_test
ea_test
root@dev1:~ # dmesg | grep epair7a
epair7a: Ethernet address: 02:0f:01:12:f1:0a
epair7a: link state changed to UP
epair7a: changing name to 'ea_test'
```

As you can see we still can grep the epair number, and this would solve our problem while keeping current code.

A second option would be to simply call "ifconfig epair create" when we create the jail, then destroy but retain the value for the jail.conf file.

I understand this doesn't do it dynamically still, but I think this PR is a bit hacky with merging 3 or 4 commands on one line.

nostrad:
right, I still hold that looking at ifconfig/dmesg/... etc cannot win in that case, you cannot see into the future.

nostrad:
again, same commands as per above, replace pkg install with
```
root@freebsd:~ #  fetch https://github.com/BastilleBSD/bastille/archive/refs/tags/0.14.20250420.tar.gz
root@freebsd:~ # tar -xzf 0.14.20250420.tar.gz 
root@freebsd:~ # cd bastille-0.14.20250420/
root@freebsd:~/bastille-0.14.20250420 # make install
```

But the rest is exactly the same and I get the same error (and yes, the VM disk image was complettly recreated before retesting)

tschettervictor:
If I'm not mistaken it is not even in 0.14 yet. You'll need to try with cloning the repo and "make install"

nostrad:
> Working perfectly for me. I can use -B and -V and each time it increments the epairs by 1.

Have you done stop on the first jail before creating the second? Because I swear this cannot be working even on master.

nostrad:
So what do we do here, I cannot for my life get any version of bastille to work, and as you cannot reproduce those failures I am hitting my head against a wall here. My efforts might be better spent migrating to Linux in the hopes of a better ecosystem...

I have one final suggestion, make unique_bridge and jib be the same codepath: https://github.com/nostrad/bastille/commit/f998566911500d2125e0aa75f9c1a13511d74ddb?w=1

I have not tried the above patch yet, but what it practically does is make a command that acts exactly like jib, with the exception if the "create the bridge" bit and thus bastille can use the exact same logic. My hope that that is interesting in the name of simplification.

nostrad:
Hang on @tschettervictor, are you running netgraph in your config, because that would explain why things work for you....

tschettervictor:
I'm not using Netgraph.

bmac2:
@nostrad I am still confused.  WHY are you renaming epairs the way you are?  I still don't follow as to why.

nostrad:
Sigh, use jib a few times outside bastille and report how many epair numbers you get, exactly none unless you explicitly do "jib eNa_something" and manually number them, because, like the -V example shows, jib does not do epair numbers, it does interface numbers:

- jib jailAnet igb0 igb1
- jib jailBnet igb0 igb1

Will create 4 epairs, named, wait for it,
e0a_jailAnet
e1a_jailAnet
e0a_jailBnet
e1a_jailBnet

It is not reasonable for a user to make a numbering scheme that just so happens to include the real epair numbering here.

tschettervictor:
@nostrad 

#1132 does what you want. 

If we decide to go forward with this, it will introduce some breaking changes, but I think it's worth it in the long run.

Can you test it?

PR #1191, opened 2025-07-16
common: Add option to use “tags” as TARGET

tschettervictor:
This will allow users to use any tag as the TARGET.


cedwards:
I have tested the branch. So far so good. Thank you.
Looks safe to merge.

PR #1210, opened 2025-08-09
Wrap $RDR_IF in double quotes

dougmacnerland:
My pf.conf looks like this, to support multiple interfaces later on:
image

When $RDR_IF is not quoted, the resulting validate_rdr_rule, persist_rdr_rule, load_rdr_rule calls will interpret $RDR_IF as 3 separate strings "{", "em0" and "}" when it should be the single string "{ em0 }"

I added some echo statements while troubleshooting to see the values of the args during `bastille rdr` command execution. Here is the failing, split $RDR_IF:
image

and here is the passing, double quoted $RDR_IF:
image


tschettervictor:
Thanks!

Issue #1215, opened 2025-08-25
[BUG] "bastille network -V" doesn't check if interface name has a dot (.) in it.


subnetspider:
**[MANDATORY] Describe the bug [MANDATORY]**
When adding a VNET interface to a jail with `bastille network -V`, there is no check of a dot (.) in the interface name. When trying to start the VNET jail, it fails.

**[MANDATORY] Bastille and FreeBSD version (paste ``bastille -v && freebsd-version -kru`` output)**
1.0.1.250714
14.3-RELEASE-p2
14.3-RELEASE-p2
14.3-RELEASE-p2

**[MANDATORY] How did you install bastille? (port/pkg/git)**
pkg

**[optional] Steps to reproduce?**
1. Create a VLAN interface on the FreeBSD host (e.g. `re0.11`).
2. Add a second interface to an existing VNEt jail: `doas bastille network -a -M -V -n jail01 add re0.11`
3. The jail failes to start after the interface is created.

**[optional] Expected behavior**
I would expect `bastille network -V` to display an error like `bastille create -V`:
```
[ERROR]: [-V|--vnet] does not support dots (.) in interface names.
```

**[optional] Screenshots**
```shell
admin@bastille02:~ % doas bastille network -a -M -V -n jail01 add re0.11

[jail01]:
jail01: removed

[jail01]:

Valid: (re0.11).
ifconfig_ng0_jail01_name: vnet0 -> vnet1
Added VNET interface: "re0.11"

[jail01]:
ng0_jail01
jail01: created
ifconfig: interface ng0_jail01 does not exist
jail: jail01: /sbin/ifconfig ng0_jail01 vnet jail01: failed
jail01: removed
```

```
# Trying to start another Jail after removing the epair and adding an ng_eiface:
admin@bastille02:~ % doas bastille start zabbix02

[zabbix02]:
ifconfig: interface re0_11 does not exist
jail: zabbix02: jng bridge zabbix02 re0_11: failed
```

**[optional]  Additional context**
This particular bastille host uses Netgraph, but it should also apply to "normal" epair + bridge setups.

tschettervictor:
Can you test master? I think we fixed this already.

subnetspider:
I just tested with the version I compiled from `https://github.com/BastilleBSD/bastille.git`, but the behavior is the same:

```
admin@bastille02 % doas bastille network -a -M -V -n jail01 add re0.10

[jail01]:
jail01: removed

[jail01]:

[...]

PR #1216, opened 2025-08-25
network: don't allow dots in interface name if -V

tschettervictor:
@subnetspider Can you test this in light of #1215


subnetspider:
It works, now `bastille network` throws an error and stops right away. :)

```
admin@bastille02:~ % doas bastille network -a -M -V -n jail01 add re0.10

[jail01]:

Valid: (re0.10).
[ERROR]: [-V|--vnet] does not support dots (.) in interface names.
```

PR #1218, opened 2025-08-27
Update man page

metalefty:
Let's put decompressed man page in git repo because the compressed man page is difficult to edit and also not suitable for tracking in version control system. 

Here are changes on the man page:
- Add edit subcommand
- Remove `bastille_list` as it is no longer work with bastille 1.0
- Remove trailing whitespace
- Update date

I think it would be very nice to mention in the document, such as "Breaking changes and migration to 1.0"

For example: 

----
The `bastille_list` parameter in `rc.conf`, present in previous versions, is no longer available. Since Bastille 1.0, all jails start automatically at boot. To disable boot-time startup, you must turn off the `boot` switch for each jail individually.

```
bastille config JAILNAME set boot off 
```

Alternatively, you can disable boot-time startup when creating a new jail.

```
bastille create --no-boot JAILNAME RELEASE IP
```

----


metalefty:
I would say there should be a separate wiki page, then I can add a link from FreeBSD's UPDATING entry.

tschettervictor:
I'm with you on the decompressed man page.

As for a wiki, do you mean something like https://bastillebsd.readthedocs.io?

tschettervictor:
> I'm thinking of a GitHub wiki page, then I can also add a note when I discover something that needs attention.
> 
> 
> 
> But readthedocs.io is also fine if it's convenient for you. Anyway, breaking changes and important notes regarding the migration from previous versions to 1.0 should be consolidated on a single page.

We've tried to do this by placing those notes on the main page of the repo.

tschettervictor (approved):

PR #1259, opened 2025-10-04
persist, eval quotes on rdr $if when necessary

dougmacnerland:
Previously, after restarting a jail on a system with support for multiple interfaces, `bastille rdr` does not create the persisted redirects. Num arg debug statements have been added:
Screenshot from 2025-09-28 13-11-29
Screenshot from 2025-09-28 13-11-55
Screenshot from 2025-09-28 13-12-32
Screenshot from 2025-09-28 13-12-47
Screenshot from 2025-09-28 13-13-02
Screenshot from 2025-09-28 13-13-49

rdr.sh changes:
Changes to persist functions wrap the `$if` arg in quotes when a space is detected:

Screenshot from 2025-10-04 09-21-33
Screenshot from 2025-10-04 09-21-19
Screenshot from 2025-10-04 09-20-59

start.sh changes:
From testing, Bourne shell's `read` built-in does not perform word-splitting correctly when the output is passed directly into the `bastille rdr` command and fails to parse quoted interface `"{ em0 }"` into a single arg, instead creating multiple args `""{"`, `"em0"`, `"}""` This has been fixed by passing the command, jail name, and rdr args into the `eval` built-in, which does its own word-splitting

### DIFF
--- a/usr/local/share/bastille/rdr.sh
+++ b/usr/local/share/bastille/rdr.sh
@@ -144,6 +144,10 @@ persist_rdr_rule() {
     local host_port="${6}"
     local jail_port="${7}"
 
+    if echo "$if" | grep -qs " "; then
+        if=\"$if\"
+    fi
+
     if ! grep -qs "$inet $if $src $dst $proto $host_port $jail_port" "${bastille_jailsdir}/${TARGET}/rdr.conf"; then
         echo "$inet $if $src $dst $proto $host_port $jail_port" >> "${bastille_jailsdir}/${TARGET}/rdr.conf"
     fi
@@ -160,6 +164,10 @@ persist_rdr_log_rule() {
     local jail_port="${7}"
     shift 7;
     log=$@;
+
+    if echo "$if" | grep -qs " "; then
+        if=\"$if\"
+    fi
 
     if ! grep -qs "$inet $if $src $dst $proto $host_port $jail_port $log" "${bastille_jailsdir}/${TARGET}/rdr.conf"; then
         echo "$inet $if $src $dst $proto $host_port $jail_port $log" >> "${bastille_jailsdir}/${TARGET}/rdr.conf"
--- a/usr/local/share/bastille/start.sh
+++ b/usr/local/share/bastille/start.sh
@@ -211,7 +211,7 @@ for _jail in ${JAILS}; do
     # Add rdr rules
     if [ -s "${bastille_jailsdir}/${_jail}/rdr.conf" ]; then
         while read _rules; do
-            bastille rdr ${_jail} ${_rules}
+            eval "bastille rdr ${_jail} ${_rules}"
         done < "${bastille_jailsdir}/${_jail}/rdr.conf"
     fi
 


tschettervictor:
Ah so you want to do something like `bastille rdr -i if1,if2,if3 TARGET...`?

dougmacnerland:
Yup, something like that

tschettervictor:
Can you test #1260 
It now allows you to specify multiple interfaces by separating them with a comma.

`bastille rdr -i vtnet0,em0 TARGET tcp 80 80` should apply the rule on both interfaces.

tschettervictor:
If you like it, I will update the docs for it also.

dougmacnerland:
I don't like this solution, because it requires the application setting up the redirect to have knowledge of the firewall settings, parse the interface-group and strip out any spaces. I enjoy using bastille for setting up containers because it handles things like that for me, and all I need to know is "I want this port on the host to map to that port on the jail" why can't bastille accept the pf grammar for interface specifications if it relies on using pf for redirects?

The real problem I was hoping to solve is restoring my redirects after a jail restart. The current rdr.conf format does not persist all of the information about the interface. If this PR is unacceptable, would a better solution be to reparse the $ext_if variable at jail start time, instead of reading it from rdr.conf?

tschettervictor:
I don't follow.

The redirect information is stored in `rdr.conf` and is loaded just fine (in my testing) on each jail restart.

tschettervictor:
> Right, but it still requires me to manually set the interface name, where previously bastille would parse it from pf.conf. 
> 
> 
> 
> Sorry if I caused some confusion here, but I don't want to specify interfaces when creating a redirect, I would prefer having bastille parse this for me and do the work it needs to do. 
> 
> 
> 
> Bastille already does this with the initial call to `bastille rdr <target> tcp 5000 5000` I'm just looking for a solution that will keep the redirect after the jail restarts.

Let me try and follow.

1. It already does allow you to `bastille rdr TARGET tcp 5000 5000`

2. It will store the rule with the currently assigned "ext_if", but sets it statically.

3. You want to change the "ext_if" inside pf.conf and have it pick that up from rdr.conf


So essentially, you want to have your rdr.conf look like the following (with the $ext_if being dynamic)

`dual $ext_if any any 5000 5000`

Am I following properly?

dougmacnerland:
Yes, but I'm only proposing this as a solution if the one proposed in this PR is unacceptable. What is wrong with wrapping the interface name in quotes when it's written to rdr.conf?

tschettervictor:
> Yes, but I'm only proposing this as a solution if the one proposed in this PR is unacceptable. What is wrong with wrapping the interface name in quotes when it's written to rdr.conf?

I guess I just can't see why. It has the exact same effect right now as I see it. The interface is still statically set inside rdr.conf

What is the main difference between what you are suggesting, and #1260?

tschettervictor:
Reading over this PR, and what I'm getting is that you want to persist the redirect after a jail restart?

This is already the behavior. When a jail is restarted, any rules inside `rdr.conf` will automatically load.

As long as you initially ran `bastille rdr TARGET tcp 5000 5000`, you should be ok. This is only run one time.

tschettervictor:
Also, what version of Bastille are you running?

tschettervictor:
Ok so you should update to 1.1.0, as I think some rdr related things have been fixed in there.

Also, I don't understand why you would want to include the brackets. Isn't it enough to just specify "em0" without them? 

Or do you have some interfaces like "em 0" (which aren't supported if I recall in FreeBSD anyway)

I don't have anything against your PR except I can't understand why you would want to include the brackets, instead of just `em0` or for multiple `em0,em1` etc...

tschettervictor:
> My goal is to make jail redirects as seamless as possible, and I don't want to specify comma separated interfaces each time I need a redirect.
> 
> 
> 
> Requiring admins to specify their interfaces like this makes automating jail deployment difficult. What about bastillefile templates? How could the author of a template know what interfaces to specify? Would the sysadmin then have to edit each template they want to install and add the appropriate interfaces? Customizing template files in this way runs counter to jail automation, and makes my life harder. 
> 
> 
> 
> I like that bastille doesn't need me to specify my interfaces every time I need to redirect. Manually specifying interfaces for each redirect is not satisfactory.

I agree 100%

But let's start from the top.

I want to rdr, so I do `bastille rdr TARGET tcp 8000 8000` which adds the entry to rdr.conf and applies it on each jail restart.
The interface it grabs is specified inside pf.conf

Now you take it from here and detail what next steps you would like to take next for any additional redirects you want to do, and how this PR would help you achieve that

Keep in mind that
1. The rule is added
2. The rule is applied on each restart

dougmacnerland:
Finally got around to testing this, everything works! Thanks @tschettervictor for all your help! 

Patch on master works, closing this PR

PR #1458, opened 2026-01-20
network: filter out comments from rc.conf

tschettervictor:


tschettervictor:
There are a handful of places like clone.sh but there it doesn't seem important. They are looking for exact matches in those cases.

This is an edge case in network.sh that almost doesn't have to be there. We are checking to see if the numbers on vnetX and eXa match. Not strictly necessary, but good to keep in line with each other.

Issue #1486, opened 2026-02-25
[BUG] cannot clone a jail on encrypted ZFS


thenktor:
**[MANDATORY] Describe the bug [MANDATORY]**
Cloning a jail on enrypted ZFS fails:
```txt
# bastille clone -a mox mox-clone 10.0.0.36

Attempting to clone 'mox' to 'mox-clone'...

Valid IP: 10.0.0.36/24

[mox]:
mox: removed
cannot send tank/encrypted/bastille/jails/mox@bastille_clone_2026-02-25-134115: encrypted dataset tank/encrypted/bastille/jails/mox may not be sent with properties without the raw flag
warning: cannot send 'tank/encrypted/bastille/jails/mox@bastille_clone_2026-02-25-134115': backup failed
cannot receive: failed to read from stream
cannot open 'tank/encrypted/bastille/jails/mox-clone/root': dataset does not exist
cannot open 'tank/encrypted/bastille/jails/mox-clone': dataset does not exist
grep: /usr/local/bastille/jails/mox-clone/jail.conf: No such file or directory
Error: Failed to update fstab:

Cloned 'mox' to 'mox-clone' successfully.

[mox]:
e0a_mox
e0b_mox
mox: created
Jail not found "mox-clone"
```

**[MANDATORY] Bastille and FreeBSD version (paste ``bastille -v && freebsd-version -kru`` output)**
1.3.2.251225
15.0-RELEASE-p3

**[MANDATORY] How did you install bastille? (port/pkg/git)**
pkg

**[optional]  Additional context**
The `--raw` argument is missing for encrypted datasets. May also be related to #839

PR #1490, opened 2026-03-03
Fix for encrypted datasets

thenktor:
As noted in #1486


thenktor:
What happened to this fix? It's not in current master and release?

tschettervictor:
Good question. I'll see if I can track it down, and re add it along with the export one.

PR #1528, opened 2026-04-12
Add support for jail to ip resolution using local-unbound

tschettervictor:
FEAT: Integrate `local-unbound` to allow resolving jail names to IPs.

To get started with this PR, add the following variables to you `bastille.conf` file.

```
## Bastille DNS
bastille_dns_enable=""
bastille_dns_interface="bastille0"
bastille_dns_gateway="10.0.0.1"
bastille_dns_domain="bastille"
```
Then run `bastille setup dns`.

This will:
1. make sure local-unbound is present
2. make sure local-unbound is enabled
3. run `local-unbound-setup`
4. copy `bastille.conf` to `/var/unbound/conf.d/bastille.conf`
5. start the services

After starting the local-unbound service, it will be necessary to restart your jails so they can be added to the resolvers list.

After restarting, change each jails resolver inside `/etc/resolv.conf` to be `nameserver 10.0.0.1` or whatever you decided to enter in the `bastille_dns_gateway` field above. For NAT jails, this should be the lookback address. For other types of jails this obviously won't work as they will have to be able to reach that IP. For example, jails attached to a bridge using epair interfaces definitely won't be able to use this as their network is isolated.
This PR is strictly for NAT jails.

After that, your host and jails should be able to communicate with each other using `jail.bastille` as a hostname. Normal outgoing traffic should also be able to fully resolve and work.

Any jails created after enabling DNS will have `bastille_dns_gateway` assigned automatically inside `/etc/resolv.conf` as the only resolver.

Notes:
Running `local-unbound-setup` will override your `etc/resolv.conf` to point to `127.0.0.1` and add whatever your nameserver was to its forwarding list.

The `bastille_dns_interface` should be the same value as the `bastille_network_loopback` and it is what `local-unbound` will listen on. Thats why NAT jails will be able to use it.

The `bastille_dns_gateway` should be set to a valid RFC1918 address. It defaults to 10.0.0.1. If you change it, just make sure to do so before running the `bastille setup dns` command.

The `bastille_dns_domain` is simply the TLD part. It can be changed to whatever you prefer. If left blank, it will default to `bastille`.


usr/local/etc/bastille/bastille.conf.sample
```diff
@@ -73,13 +73,22 @@ bastille_export_options=""                                            ## default
 
 ## Networking
 bastille_network_vnet_type="if_bridge"                                ## default: "if_bridge"
-bastille_network_loopback="bastille0"                                 ## default: "bastille0"
 bastille_network_pf_ext_if="ext_if"                                   ## default: "ext_if"
 bastille_network_pf_table="jails"                                     ## default: "jails"
 bastille_network_shared=""                                            ## default: ""
+bastille_network_loopback="bastille0"                                 ## default: "bastille0"
+bastille_network_shared=""                                            ## default: ""
```
michael-o, line 80: This is now duplicated.

michael-o (changes requested):

PR #1529, opened 2026-04-13
monitor: add uptimekuma monitor module

tschettervictor:
Please make sure to diff the config file as this changes an important variable in the monitoring configuration.


tschettervictor:
@diktomat Im pinging you because you're the only one I know right now that probably uses this.

PR #1554, opened 2026-06-06
Documentation formatting fixes

djberg96:
Just some doc fixes for issues I ran into while building a PDF.

I'm on MacOS 26.5.1, and built with `make clean latexpdf SPHINXBUILD=../.venv-docs/bin/sphinx-build` in case that matters.

Mainly it's missing mandatory newlines after directives and some markdown that got mixed in by mistake I assume, along with a couple missing things. The changes to conf.py were mainly to eliminate some build warnings, but it also adds in the logo on the cover page so it looks nice.


tschettervictor:
Great!

PR #1565, opened 2026-06-27
Secure RDR by splitting the dynamic rules into rdr+filter

tschettervictor:
Previous to this, rdr rules would always bypass all block rules because of the sturcture of the pf.conf file.

This PR addresses this by splitting it into two parts. The rdr anchor now simply handle the translation of the address, and the filtering anchor handles the actual passing of traffic. This should allow block rules to work properly.

### DIFF
--- a/docs/chapters/networking.rst
+++ b/docs/chapters/networking.rst
@@ -607,6 +607,8 @@ Create the firewall rules:
   block in all
   pass out quick keep state
   antispoof for $ext_if inet
+
+  anchor "bastille/*"
   pass in proto tcp from any to any port ssh flags S/SA modulate state
 
 - Make sure to change the ``ext_if`` variable to match your host system
--- a/docs/chapters/subcommands/rdr.rst
+++ b/docs/chapters/subcommands/rdr.rst
@@ -3,7 +3,8 @@ rdr
 
 ``bastille rdr`` allows you to configure dynamic rdr rules for your containers
 without modifying pf.conf (assuming you are using the ``bastille0`` interface
-for a private network and have enabled ``rdr-anchor 'rdr/*'`` in /etc/pf.conf as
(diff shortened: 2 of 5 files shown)


laktech:
Whats' the difference between the newly introduced `bastille/*` anchor and the anchor configured in bastille.conf under the configure key `bastille_network_pf_table`? Why isn't `bastille/*` anchor name configurable?

Some jails have anchors mounted under the `bastille_network_pf_table` anchor and some under the newly introduce bastille anchor while all rdr rules remain in the `rdr/*` -- what excatly was achieved here?

tschettervictor:
The `bastille_network_pf_table` is only there so jails can have outbound access.

In the old version, we used `rdr pass...` for the `rdr-anchor` rules, which would automatically pass any traffic that came in on the port specified. If you added some block rules in the PF filtering section, they would not apply to the rdr, since the traffic was already passed through.

By splitting it into two anchors, we can now perform the redirection, and apply additional firewall rules.

For example, if i perform a redirection for port 443 into my jail, and then add a filtering rules in the filtering section using a block table (for example) then IPs that are blacklisted inside the filtering table will not be able to reach my jail because of the new filtering anchor. 

Before this change, IPs could still get into the jail, becaue of the `pass` keyword in the `rdr-anchor`.

laktech:
I see thanks for that explaintation. I dug into my setup some more and it seems like I was confused by what came out of the box w/ bastille and what was my own invention.

`bastille_network_pf_table` is just the table--not an anchor. In creating my own work-around for the use-case this PR addresses. I created the `jails` anchor and attached firewall rules to it using prestart / poststop hooks with a pf_rules.conf file that defines the firewalls rules and lives in the per-jail directory. I wonder if I can simplify this now that this capabliity exists.

I'd have to dig into this some more to see if/how bastille supports declaring firewall rules and having those attach to the `bastille` anchor.

PR #1595, opened 2026-07-17
feat(vm): add bhyve VM support as a jail peer (RFC/draft)

usenix17:
## RFC / Draft: first-class bhyve VM support

**This is a draft opened as an RFC -- I'm seeking design feedback, not asking for a merge yet.** The manifest format is a long-term commitment, so I'd rather agree on the shape with maintainers before polishing.

### Summary

Adds bhyve virtual machines as a peer instance type to jails. A VM is defined by a git-trackable template, provisioned with the existing template workflow, and driven by the same lifecycle verbs (`create`/`start`/`stop`/`console`/`list`/`destroy`). The goal is a single management plane for every workload on a FreeBSD host -- whether it runs on the host kernel (a jail) or brings its own (a VM). Deliberately inspired by illumos zone brands and SmartOS ("everything is a zone, the brand is a detail").

### What's here

- New `bhyve.sh` library: template parser, authoritative `vm.conf`, zvol-backed disks, tap-on-bridge networking, bhyve(8) argument generation (materialized to `bhyve.args`), a supervision jail with `allow.vmm`, nmdm(4) serial console, and rctl memory capping.
- Brand dispatch across `create`/`start`/`stop`/`console`/`destroy`/`restart`/`list` -- the jail path is untouched when the target isn't a VM.
- New Bastillefile verbs in VM mode: `VM CPU MEM BOOTROM DISK NIC ISO ADDRESS` (jail-only verbs error in VM mode; a template is one brand).
- `bastille create --vm NAME TEMPLATE`; `bastille list` grows `vm` rows via the existing Type column plus a `list vms` subcommand.
- Docs chapter, `bastille-create(8)` update, example template, and rc.d boot/shutdown ordering.

### Validated on bare-metal FreeBSD 15.0-RELEASE amd64

An Alpine guest boots to a serial login over nmdm, appears in `bastille list` as a `Type=vm` peer with a real JID, reboots correctly (bhyve relaunches rather than powering off), stops gracefully (ACPI power button, then forced after a timeout), and destroys cleanly -- with the host `/dev`/`zfs` unaffected throughout. ShellCheck clean at this repo's CI severity and exclusions.

### Scope (deliberately minimal)

Non-goals for v1: suspend/resume, live migration, cloud-init/ignition, VNC/graphical console, Windows guests, PCI passthrough, multi-host orchestration. RDR-to-pf wiring for VMs is recorded but deferred. ZFS is required.

### Open questions for maintainers

1. **In Bastille or beside it?** This is the first conversation to have. The design works either way, but the manifest format deserves maintainer buy-in even if the code starts external.
2. ~~**Supervision-jail devfs confinement.** v1 uses `path=/` and shares the host `/dev`. A restrictive per-VM devfs ruleset (exposing only the VM's vmm/tap/nmdm devices) needs a dedicated jail root with an isolated `/dev`, and is deferred -- because `path=/` + `mount.devfs` + a ruleset stacks a locked-down devfs over the host's own `/dev` and hides host devices like `/dev/zfs`. Left out until it can be done with an isolated root.~~ **Update: done.** The supervision jail now runs in its own root (a read-only nullfs skeleton of only the userland bhyve needs) with a private per-VM devfs that exposes only that VM's own nodes (`vmm`/`vmm.io`/`nmdm`/tap/zvols plus `null`/`zero`/`random`/`pci`) and hides `/dev/mem`, `/dev/kmem`, raw host disks, and every other VM. It also gets its own vnet (empty in shared mode, full in VNET mode), so it sees no host interfaces. On by default via `bastille_vm_isolated_devfs`, validated on bare metal in both networking modes.
3. **Manifest stability** (a `version=` key from day one?), **naming** (`--vm` flag vs a `bastille vm` subcommand), and **rctl defaults** -- all discussed in the design.

I'm happy to open a Discussion or attach the full design document (motivation, prior art -- vm-bhyve / CBSD / illumos zones / kernel zones, risks and mitigations) if that's the preferred venue for the design conversation.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -29,6 +29,7 @@ Christer Edwards [christer.edwards@gmail.com]
 - Eric Borisch
 - Kevet Duncombe
 - Victor Tschetter
+- Sasha Karcz
 
 ### Special thanks
 Software doesn't happen in a vacuum. Thank you to the following people who may
--- a/docs/chapters/vm.rst
+++ b/docs/chapters/vm.rst
@@ -54,8 +54,8 @@ Verb reference
 - ``BOOTROM spec`` -- ``uefi`` (default), ``uefi-csm``, or a firmware path.
 - ``DISK name size [zfsprop=val ...]`` -- creates and attaches a virtio-blk
   zvol in declaration order.
-- ``NIC bridge`` -- creates a persistently named tap (``vm-<name>-<n>``) and
-  adds it to the named bridge (default: ``bastille_vm_bridge``).
+- ``NIC bridge`` attaches the guest NIC to the named bridge (default:
+  ``bastille_vm_bridge``). See "Networking modes" below.
 - ``ISO url|path`` -- attaches installation media as an ahci-cd device. Remote
   URLs are fetched into the Bastille cache.
 - ``ADDRESS ip`` -- records the guest address (required for ``RDR``).
@@ -64,6 +64,21 @@ Verb reference
 
 Jail-only verbs (``PKG``, ``SYSRC``, ``CP``, ``CMD``, ``SERVICE`` ...) are
 errors in a VM template; a template is one brand.
+
+Networking modes
+~~~~~~~~~~~~~~~~
+
+A VM's networking has two modes, chosen at create time the way jails choose
+theirs:
+
+- **shared** (default): the guest's tap is a member of the named host bridge,
+  the pattern vm-bhyve uses. Simple, and the guest address is configured inside
+  the guest.
+- **VNET** (``bastille create --vm -V``): the supervision jail becomes a VNET
+  jail, and the guest's tap lives inside the jail's own network stack, uplinked
+  to the host bridge through an epair. The host sees only the a-side of the
+  epair, so the VM's networking is fully contained in its jail, exactly like a
+  VNET jail. This is the mode that makes "a VM is a jail" complete.
 
 Lifecycle
 ---------
--- a/docs/chapters/vm.rst
+++ b/docs/chapters/vm.rst
@@ -52,8 +52,13 @@ Verb reference
 - ``CPU n`` -- virtual CPU count (``bhyve -c``).
 - ``MEM size`` -- guest memory, e.g. ``8G`` (``bhyve -m``).
 - ``BOOTROM spec`` -- ``uefi`` (default), ``uefi-csm``, or a firmware path.
-- ``DISK name size [zfsprop=val ...]`` -- creates and attaches a virtio-blk
-  zvol in declaration order.
+- ``DISK name size [source=image] [zfsprop=val ...]`` creates and attaches a
+  virtio-blk zvol in declaration order. With ``source=<url or path>`` Bastille
+  populates the disk from a cloud/disk image at create time: a raw image is
+  written with ``dd`` (no dependency), a qcow2 image is converted straight onto
+  the zvol with ``qemu-img`` (``sysutils/qemu-tools``). This is how you boot a
+  pre-built cloud image and let cloud-init provision it, instead of installing
+  from an ISO.
 - ``NIC bridge`` attaches the guest NIC to the named bridge (default:
   ``bastille_vm_bridge``). See "Networking modes" below.
 - ``ISO url|path`` -- attaches installation media as an ahci-cd device. Remote
@@ -64,11 +69,17 @@ Verb reference
   (``alpine-virt-3.21.iso`` becomes ``Alpine-3.21``). Falls back to
   ``uefi-guest``.
 - ``CLOUDINIT file`` takes a path (relative to the template directory, or
-  absolute) to a cloud-init user-data file. Bastille builds a NoCloud ``CIDATA`` seed ISO
-  from it (plus an auto-generated meta-data with the VM's instance-id and
-  hostname) and attaches it as a CD. A cloud-init-enabled guest applies it at
-  first boot: users, SSH keys, packages, ``runcmd``, network, and so on. Uses
-  base ``makefs(8)``, no ports dependency.
+  absolute) to a cloud-init user-data file. Bastille builds a NoCloud ``CIDATA``
+  seed from it (plus an auto-generated meta-data with the VM's instance-id and
+  hostname) and attaches it as a virtio-blk disk. A cloud-init-enabled guest
+  applies it at first boot: users, SSH keys, packages, ``runcmd``, and so on.
+  Uses base ``makefs(8)``, no ports dependency. The seed is virtio-blk (not an
+  optical device) so it is visible when the guest's cloud-init ``ds-identify``
+  runs early in boot.
+- ``NETWORK_CONFIG file`` adds a cloud-init network-config (v1 or v2) to the
+  seed, so the guest gets a deterministic (for example static) address at first
+  boot. Renderer-agnostic: cloud-init applies it as netplan on Ubuntu or
+  systemd-networkd on Debian.
 - ``RDR proto host_port vm_port`` -- recorded for a future release; pf wiring
   for VMs is not yet enabled.
 
--- a/docs/chapters/vm.rst
+++ b/docs/chapters/vm.rst
@@ -63,6 +63,12 @@ Verb reference
   (e.g. ``Ubuntu 24.04``). Optional; if omitted it is guessed from the ISO name
   (``alpine-virt-3.21.iso`` becomes ``Alpine-3.21``). Falls back to
   ``uefi-guest``.
+- ``CLOUDINIT file`` takes a path (relative to the template directory, or
+  absolute) to a cloud-init user-data file. Bastille builds a NoCloud ``CIDATA`` seed ISO
+  from it (plus an auto-generated meta-data with the VM's instance-id and
+  hostname) and attaches it as a CD. A cloud-init-enabled guest applies it at
+  first boot: users, SSH keys, packages, ``runcmd``, network, and so on. Uses
+  base ``makefs(8)``, no ports dependency.
 - ``RDR proto host_port vm_port`` -- recorded for a future release; pf wiring
   for VMs is not yet enabled.
 
--- a/docs/chapters/vm.rst
+++ b/docs/chapters/vm.rst
@@ -78,8 +78,15 @@ Verb reference
   runs early in boot.
 - ``NETWORK_CONFIG file`` adds a cloud-init network-config (v1 or v2) to the
   seed, so the guest gets a deterministic (for example static) address at first
-  boot. Renderer-agnostic: cloud-init applies it as netplan on Ubuntu or
-  systemd-networkd on Debian.
+  boot. cloud-init renders it to whatever the guest uses: netplan or
+  systemd-networkd on Debian/Ubuntu, ifupdown (``/etc/network/interfaces``) on
+  Alpine. One portability caveat: the netplan-style ``match`` selector is only
+  honored by the netplan/networkd renderers. The ifupdown renderer uses the
+  ``ethernets`` key as the literal interface name, so for those guests name the
+  device directly (``eth0``) instead of a logical name plus ``match``. For
+  example, ``ethernets: {eth0: {addresses: [...]}}`` works everywhere, whereas
+  ``ethernets: {primary: {match: {name: "e*"}, ...}}`` only works on
+  netplan/networkd guests.
 - ``RDR proto host_port vm_port`` -- recorded for a future release; pf wiring
   for VMs is not yet enabled.
 
--- a/docs/chapters/vm.rst
+++ b/docs/chapters/vm.rst
@@ -1,0 +1,121 @@
+Virtual Machines (bhyve)
+========================
+
+Bastille can manage `bhyve(8)` virtual machines as a peer instance type
+alongside jails. A VM is defined by a git-trackable template, provisioned with
+the existing template workflow, and driven by the same lifecycle verbs as
+jails: ``create``, ``start``, ``stop``, ``console``, ``list`` and ``destroy``.
+
+The goal is a single management plane for every workload on a FreeBSD host,
+whether it runs on the host kernel (a jail) or brings its own kernel (a VM).
+
+Requirements
+------------
+
+VM support is amd64-only and requires hardware virtualization (VT-x/EPT). On
+the host:
+
+.. code-block:: shell
+
+  pkg install edk2-bhyve
+  kldload vmm nmdm if_bridge if_tap
+
+VMs require ZFS. Set ``bastille_zfs_enable=YES`` and ``bastille_zfs_zpool`` in
+``bastille.conf``. See the ``VM (bhyve) options`` block in the sample config
+for the tunables (firmware path, default bridge, shutdown timeout, disk/NIC
+device models, and the supervision-jail devfs ruleset).
+
+The VM template
+---------------
+
+A VM template is an ordinary Bastille template whose ``Bastillefile`` opens
+with the ``VM`` verb. The verbs after it configure the guest rather than a
+jail:
+
+.. code-block:: shell
+
+  # Template: sasha/forgejo-runner-vm
+  VM
+  CPU 4
+  MEM 8G
+  BOOTROM uefi                 # uefi | uefi-csm | path to firmware
+  DISK disk0 40G               # name + size; a zvol is created on create
+  DISK disk1 200G volmode=dev  # extra disks; zfs props pass through
+  NIC bridge0                  # a tap is created and added to bridge0
+  ISO https://.../rocky-9.iso  # installation media (first boot)
+  ADDRESS 10.0.0.40            # guest address (used by RDR)
+
+Verb reference
+~~~~~~~~~~~~~~
+
+- ``VM`` -- switches the template into VM mode. Must be the first verb.
+- ``CPU n`` -- virtual CPU count (``bhyve -c``).
+- ``MEM size`` -- guest memory, e.g. ``8G`` (``bhyve -m``).
+- ``BOOTROM spec`` -- ``uefi`` (default), ``uefi-csm``, or a firmware path.
+- ``DISK name size [zfsprop=val ...]`` -- creates and attaches a virtio-blk
+  zvol in declaration order.
+- ``NIC bridge`` -- creates a persistently named tap (``vm-<name>-<n>``) and
+  adds it to the named bridge (default: ``bastille_vm_bridge``).
+- ``ISO url|path`` -- attaches installation media as an ahci-cd device. Remote
+  URLs are fetched into the Bastille cache.
+- ``ADDRESS ip`` -- records the guest address (required for ``RDR``).
+- ``RDR proto host_port vm_port`` -- recorded for a future release; pf wiring
+  for VMs is not yet enabled.
+
+Jail-only verbs (``PKG``, ``SYSRC``, ``CP``, ``CMD``, ``SERVICE`` ...) are
+errors in a VM template; a template is one brand.
+
+Lifecycle
+---------
+
+.. code-block:: shell
+
+  bastille create --vm builder sasha/forgejo-runner-vm
+  bastille start builder
+  bastille console builder        # serial console; detach with ~.
+  bastille list                   # VMs appear with Type 'vm'
+  bastille stop builder
+  bastille destroy builder
+
+``bastille list`` grows a VM row per VM with a ``Type`` column of ``vm``;
+``bastille list vms`` shows only VMs.
+
+How it works
+------------
+
+Each VM's ``bhyve`` process runs inside a minimal, auto-generated
+*supervision jail* with ``allow.vmm``. This gives the VM a real jail ID -- so
+``jls`` and ``rctl`` treat it as a peer of jails -- while layering jail
+confinement (namespace + rctl limits) on top of bhyve's Capsicum sandbox.
+
+.. note::
+
+   The v1 supervision jail uses ``path=/`` and shares the host's ``/dev``. A
+   restrictive per-VM devfs ruleset that exposes only the VM's own vmm/tap/nmdm
+   devices is the intended additional hardening, but it requires giving the
+   jail its own root so its ``/dev`` is isolated from the host's -- deferred to
+   a follow-up. The ``bastille_vm_devfs_ruleset`` config key is reserved for it.
+
+The instance layout mirrors a jail's:
+
+.. code-block:: text
+
+  ${bastille_vmdir}/<name>/
+    vm.conf          # canonical VM definition (edit this)
+    supervisor.conf  # generated supervision jail.conf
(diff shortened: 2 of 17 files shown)


tschettervictor:
This may be interesting, but are you aware of vm-bhyve package? It is literally the same as Bastille except for VMs.

usenix17:
> This may be interesting, but are you aware of vm-bhyve? It is literally the same as Bastille except for VMs.

Absolutely -- I'm a fan of vm-bhyve, and this design actually borrows its datastore model (a ZFS dataset per VM, config inside). I called it out as prior art. So to be clear: I'm not trying to reimplement or replace it as a bhyve manager.

The proposal isn't "add a way to manage bhyve to Bastille." It's "make a VM a *peer instance type to a jail*, under one management plane." That distinction is the whole point:

-  A VM is just a template whose Bastillefile opens with `VM`. `bastille bootstrap <git-url>` rebuilds the entire host -- jails *and* VMs -- from a single template repo. vm-bhyve is its own universe: separate config format, separate CLI, no template/bootstrap story, no pf integration.
- **A VM *is* a kind of jail.** Each bhyve process runs inside an auto-generated supervision jail with `allow.vmm`, so the VM gets a real JID -- `jls` and `rctl` treat it as a peer of jails, and rctl bounds the guest's wired RAM. That jail-confinement-over-Capsicum arrangement is tedious to plumb by hand; Bastille is the natural place to make it the default.
- **One lifecycle.** `start`/`stop`/`console`/`list`/`destroy` are brand-agnostic, and `bastille list` shows jails and VMs interleaved with a `Type` column. The jail code path is untouched when VM support is unused.

The north star is illumos/SmartOS zone brands: a bhyve guest is a zone with `brand=bhyve`, and `zoneadm list` shows containers and VMs as peers. FreeBSD has every primitive for the equivalent (jails, vmm, ZFS, pf, nmdm) -- what's missing is the integration layer, and Bastille already owns the right abstractions.

Totally fair if the conclusion is "this belongs *beside* Bastille rather than in it" -- that's genuinely the first question worth settling, which is why I opened this as a draft/RFC. Happy to share the full design doc if it's useful.

yaazkal:
@usenix17 can you please share a use case?

usenix17:
Sure -- and it's my actual host, not a hypothetical.

I run a FreeBSD box that's mostly jails: a DNS resolver, a reverse proxy, a mesh-network relay, a handful of small services -- all Bastille templates in a git repo, so `bastille bootstrap <repo>` rebuilds the whole host and `bastille list` describes it.

Then I hit workloads a jail can't hold -- Linux software that hard-requires systemd/cgroup v2 (a CI runner, a couple of Linux-only appliances). Today that means running vm-bhyve *alongside* Bastille. vm-bhyve is great, but it's a separate universe: its own config format, its own CLI, its own backup story, no shared pf/RDR, no template/bootstrap. So my host stops being reproducible from one repo, `bastille list` stops describing it, and I'm carrying two mental models for what is operationally one fleet.

What I want isn't "a way to run VMs" -- vm-bhyve does that fine. It's for the VMs and jails to be **one fleet, in one repo, under one tool**, because operationally they already are:

- one config language, one git tree; `bastille bootstrap` rebuilds jails and VMs alike
- `bastille list` shows them as peers; `start`/`stop`/`console`/`destroy` are brand-agnostic
- `zfs snapshot -r` captures the whole host -- jails and VMs -- atomically
- `RDR` uses the same pf anchors regardless of backend
- each VM runs in a thin supervision jail with `allow.vmm`, so it gets a real JID -- `jls`/`rctl` treat it as a jail peer, which makes "the brand is a detail" literal rather than a slogan

Concretely, that's a single `bastille list` with a Linux guest sitting next to the jails:

```
 JID  Name       State  Type   IP Address   Release
 3    resolver   Up     thin   10.0.0.53    15.0-RELEASE
 4    proxy      Up     thin   10.0.0.7     15.0-RELEASE
 9    ci-runner  Up     vm     10.0.0.40    uefi-guest
```

That single pane, rebuilt from one repo, is the whole use case. vm-bhyve can't give me that (it isn't Bastille), and it's the opposite of the web-appliance model (Sylve/ClonOS/Proxmox) -- which is exactly what I chose Bastille to avoid.

tschettervictor:
Does cbsd do anything similar?

So essentially it's a VM running inside a jail? All the tap/bridges would be in the jail right, with the host only seeing the a side of the epair?

Also how would this work "alongside Bastille" as you put it? I mean if the tool would be separate?

Also, what about integrating a bhyve management plane that would run VMs along side jails as opposed to running them inside jails?

JRGTH:
This is a really nice concept and I understand your use case on the management ease, also actually useful over the experimental Linux jails despite VM's being heavier on system resources, though the last may be subjective.

Between as any new POC, this opens for a lot of questions as expected.
I assume "VM" type jails is also a Thick jail, i.e. a self contained and fully independent jail with full export/import support?

Also how this will integrate/behaves with an existing `vm-bhyve`(with a single public bridge shared to Bastille) deployment with existing VM's and VNET jails?.

I assume that the "VM" type jails even with enabled `allow.vmm` the VM will also inherit some of the kernel restrictions of the jail peer affecting things such as device pass-through etc.?

Guess is time to tinker with on few alternate boot environments between.

Regards

usenix17:
Great questions. In order:

**Does CBSD do anything similar?** At the "manage jails and VMs under one roof" level, yes, and I cite it in the design as the maximal version of this idea. But as far as I know CBSD runs bhyve on the host; it does not run the device model inside a jail (you would know its internals better than I would). So the specific thing here, a VM that *is* a jail (real JID, `rctl` peer, `allow.vmm` confinement), is not what CBSD does.

**"A VM inside a jail? taps/bridges in the jail, host seeing the a-side of an epair?"** That is exactly the model I think it should be, and I tested it to be sure. The supervision jail can be a full VNET jail: the guest's tap lives inside the jail's own network stack, bridged to an epair whose other end sits on the host bridge. I stood this up on a test box, gave an Alpine guest an address, and pinged it from the host straight through the jail's vnet (0% loss). So yes: taps and bridge inside the jail, host sees only the a-side of the epair, precisely as you described. (v1 in this PR currently uses the simpler host-tap-on-bridge pattern that vm-bhyve uses, but the VNET model works and reuses Bastille's existing epair and bridge plumbing, so that is the direction.)

**"How would this work alongside Bastille if the tool is separate?"** Sorry, "alongside" was describing today's pain: running vm-bhyve (a separate tool) next to Bastille. The proposal is the opposite. This is integrated into Bastille: `bastille create --vm` / `start` / `list` / `destroy`, one CLI, no second tool.

**"What about a management plane that runs VMs alongside jails instead of inside them?"** I would push to keep them in jails, and it is the crux of why this is worth doing. Running bhyve on the host beside jails is what CBSD and vm-bhyve already do; a management plane that does the same is just another one of those. The value here is that the VM *is* a jail: it gets a real JID, `jls` and `rctl` treat it as a first-class jail, and with the VNET networking above it has its own network stack exactly like a VNET jail. That is what makes "the brand is a detail" architecturally true rather than a UI label, and it keeps the code as brand dispatch over one instance model instead of two parallel subsystems. The cost is honest: the supervision jail is moving parts, and v1's isolation is partial since it shares the host `/dev` (an isolated per-VM devfs is on the roadmap). What you get for it is a VM that is genuinely a jail rather than a process running beside them, which is the whole reason to prefer this over running bhyve on the host.

usenix17:
Thanks, glad the management-ease angle lands, and the point about it being more useful than experimental Linux jails is fair.

- **Thick / export-import:** the supervision jail is diskless (`path=/`), so thick-vs-thin does not map onto it, but the VM as a unit is self-contained: manifest in the instance dir, disks as zvols under one `vms/<name>` dataset. Export/import is therefore a `zfs send` of that dataset plus the dir, to be added later.
- **Coexisting with vm-bhyve on a shared bridge:** no conflict. A Bastille VM just adds its tap (shared mode) or epair (VNET mode) to the bridge the way vm-bhyve does, and never touches vm-bhyve's VMs or your VNET jails.
- **Jail restrictions:** yes, correct. The jail constrains the bhyve process, so `allow.vmm` gets the vmm device, but passthrough needs the `ppt` device exposed to the jail plus host prep, and a passthrough VM ends up host-pinned. That is exactly why passthrough is "supported with caveats" on the roadmap.

Please do tinker. It boots and runs on the branch today (validated on bare metal), so a throwaway BE is perfect.

tschettervictor:
Have you looked at building a template for this? This would mean installing vm-bhyve inside the jail. See the following example:

```
ARG+ NAME=myvm
ARG+ ISO
ARG+ DISK
ARG+ SIZE
ARG+ TEMPLATE

PKG vm-bhyve edk2-bhyve uefi-firmware

CONFIG allow.vmm 1
RESTART

SYSRC vm_dir="zfs:tank/data"
SYSRC vm_enable="YES"

CMD vm init
CMD vm switch create public
CMD vm switch add public vnet0
CMD vm create -t ${TEMPLATE} -s ${SIZE} ${NAME} 
CMD vm install ${NAME} ${ISO}
```

Obviously this is simple and not production based, but you could then put the template, along with a vm-bhyve template file into the Bastille template repo, then simply specify it as the vm-bhyve template for the vm to use.

Obviously we could then add some baseline functionality to detect this and list the jail type as vm in such a case, as well as making the console exec into the vm as opposed to the jail, though that might require a switch.

IMO it somewhat deviates from the original design of Bastille, though it does have some use cases. But if we can simply utilize a tool that is already very mature (vm-bhyve) then I'd opt to do that rather than reinvent the wheel.

usenix17:
Thanks, this is a fair question, and vm-bhyve's maturity is why it's tempting.

But the two approaches build different things. A `PKG vm-bhyve` + `allow.vmm` template gives you a jail that runs vm-bhyve, which spawns guests inside it. `jls`/`rctl` see one jail running a VM manager, and Bastille has to reach through vm-bhyve (its config, its `vm_dir`, its switches) to touch the guests. That's why you'd then need to detect the special jail, relabel its type, and add a console switch: it's the "brand is a UI label" version.

This PR makes each VM be a jail. `bastille create --vm` generates a supervision jail with `allow.vmm` running one bhyve device model, so the guest has a real JID, is an `rctl`/`jls` peer, and with VNET has its own network stack like any VNET jail. One instance model, one manifest language, one CLI, no second tool. And it doesn't reinvent bhyve: it orchestrates `bhyve(8)` directly (generating the argument vector), the same primitive vm-bhyve wraps. Wrapping vm-bhyve instead keeps the exact model this is trying to move past, a VM as a host process beside jails rather than a jail.

There's more investment in this design than in a template, but it's what creates a VM that is genuinely a jail rather than a VM manager hidden behind one. It boots on the branch today, with per-VM devfs and vnet isolation so the containment is real. A vm-bhyve template is still fine to have for folks who just want vm-bhyve managed by Bastille; it's just a different feature.

tschettervictor:
You also stated in the initial post in question 1, "In Bastille or beside it". What would "beside it look like"?

Also, is this PR hand written, or AI written? Thanks.

yaazkal:
Thanks @usenix17 for the explanation about the use case. What I can understand, the core value here is to add a new 'vm' jail type. For that, the PR should be as concise as possible, this PR tries to add a lot of responsibility to Bastille.

usenix17:
Completely agree, and that is exactly why this is a draft. The new `vm` jail type is the core value, and a real PR should land just that: the brand dispatch and lifecycle (`create`/`start`/`stop`/`console`/`list`/`destroy`) over one bhyve device model in a supervision jail, nothing more.

Everything layered on top here (cloud-init, clone, networking modes, the devfs/vnet hardening) is in the draft to show the whole shape and invite comment on the direction, not to land in one change. It maps to the phased plan in the design doc, and each piece would be its own small, reviewable, independently mergeable PR once the core model is agreed. I would never drop this much on the maintainers at once.

And on the maintenance burden: I am willing to own and keep supporting the code I contribute, so it is not weight the existing maintainers carry alone.

usenix17:
Two things:

On "beside it": it would mean shipping the VM support outside Bastille core rather than as a built-in brand. Concretely, a companion tool or port (say a `bastille-vm`) that uses the same manifest format and conventions but lives and versions separately, so the manifest design still gets maintainer buy-in even if the code starts external. "In Bastille" means the brand dispatch lives in Bastille itself, which is what this draft shows. I lean toward in Bastille, but the manifest format is the long-term commitment either way, so that is the piece worth agreeing on first.

On hand written vs AI: it is AI-assisted. I drove the design and the architectural decisions and validated every piece on bare metal, but a lot of the code was written with Claude and then reviewed and tested by me. Happy to go deep on any part of it, and as I said, I intend to maintain what I contribute.

tschettervictor:
I propose the following:

A template can just as easily create a VM inside a jail. I tested the following.

```
# Template to install a VM inside a jail

ARG+ NAME=vm-test
ARG+ ISO_FILE=FreeBSD-14.4-RELEASE-amd64-bootonly.iso
ARG+ ISO_URL=https://download.freebsd.org/ftp/releases/ISO-IMAGES/14.4/${ISO_URL}
ARG+ SIZE=20G
ARG+ TEMPLATE=test

PKG vm-bhyve edk2-bhyve

CONFIG set allow.vmm 1
RESTART

SYSRC vm_dir="/usr/local/vm-bhyve"
SYSRC vm_enable="YES"

CP usr /

CMD vm init
CMD vm switch create public
CMD vm switch add public vnet0
CMD vm iso ${ISO_URL}
CMD vm create -t ${TEMPLATE} -s ${SIZE} ${NAME} 
CMD vm install ${NAME} ${ISO_FILE}
```
This template has a file at `usr/local/vm-bhyve/.templates/test.conf`: 

```
loader="bhyveload"
cpu=1
memory=256M
network0_type="virtio-net"
network0_switch="public"
disk0_type="virtio-blk"
disk0_name="disk0"
```

Obviously we could use zfs if we jailed the dataset, but that might be unnecessary.

This approach works and the VM boots and can be controlled from inside the jail using vm-bhyve.

I understand that it's not as "manageable" from a single program, but it achieves what we want.

Depending on where we go here, we can also add some code to list such a setup as "thin-vm" or "thick-vm" or even just "vm". As well as allowing the console command to directly console into the VM if it is running.

Start/stop/rctl etc will still all function as they should.

usenix17:
Thanks for actually building and testing it. That is the useful kind of pushback, and it does work.

Let me concede two things up front. It is less code and reuses a mature tool. And 1:1 is not the line, you can run vm-bhyve one VM per jail if you want, so "a VM per jail" is not what separates the approaches.

Where I still think the native brand is stronger is operational, and concretely so.

1. Thin jail that shares the host's bhyve. The supervision jail is a read-only nullfs skeleton that mounts the host's `/usr/sbin` and `/usr/local/share`, so it uses the host's bhyve and edk2 firmware rather than its own copy. Patch bhyve or the firmware once on the host and every vm-type jail picks it up on next boot. The vm-bhyve template installs the hypervisor into each jail (`PKG vm-bhyve edk2-bhyve`), so you have N copies to update and they can drift. The supervision jail installs nothing, it is just mounts.

2. Per-VM isolation. The supervision jail has its own restricted devfs that exposes only that VM's nodes, and its own vnet. A vm-bhyve host jail needs broad device access, and the moment it runs more than one VM those VMs share its devfs, vnet, and rctl.

3. ZFS-native disks. Disks are zvols, so cloning a VM is a near-instant copy-on-write clone (sub-second on my test box), plus snapshot and `zfs send`. A plain `vm_dir` makes the disks files, and none of that applies.

4. One configuration language. The VM is described in the same manifest as everything else, no split brain between a Bastillefile for the jail and a vm-bhyve `.conf` for the VM.

vm-bhyve lets you run many VMs in one jail and do what you like, with Bastille only owning the jail, and that is real flexibility and minimizes code ownership on the Bastille side of the house. But it is the flexibility to give up per-VM rctl, devfs, and vnet isolation, and to accept shared fate, where one jail stop takes them all. The reason to put a VM in a jail is to get those per VM, so running many per jail argues against the reason to do it at all.

So I will grant that the vm-bhyve template approximates the goal, and it is genuinely simpler. This is the maintainers' call and I am happy either way. If the project wants that route I will not argue it into the ground, and if there is appetite for the native brand it works today and I will maintain it. I just think the approximation drops the parts that were the point: patch the hypervisor once, and every VM gets its own limits, devfs, and network stack.

tschettervictor:
So,let me take your comments one at a time:

1. We are still sharing the hosts bhyve because the jail is created as a thin jail (optional of course). That said, we might be able to add an experimental option to create a jail in the way you specify. We have done similar things in the past with Linux jails. It's a simple matter of modifying the create command to build a jail.conf that will suit our needs.

2. I'm still talking about running a single VM per jail, not multiple. So as in point 1, a jail.conf including a custom devfs ruleset would probably work here.

3. Bastille allows jailing datasets, but I haven't played with that in this context. Also, we can create the zvols outside of the jail and mount them in (optionally).

4. I hear you there.

Your summary at the end talks about one VM per jail. That is also what I have in mind with the template I built.

Also an advantage of the template approach is that all configuration stays confined to the jail. This includes taps,bridges,isofiles etc...

usenix17:
This is good, I think we are closer than it looks. On 1, 2, 3, and your one-VM-per-jail point, we now agree on the same jail: thin, sharing the host's bhyve, one VM, a custom devfs ruleset, zvols mounted in. That is the model I am after, so if the template route can build that jail, most of my concern is answered.

The one thing left is not the jail, it is the manager. In your template the guest is still driven by vm-bhyve running inside the jail (vm create, vm install, vm start). So even with the thin jail and the devfs ruleset, you still have two config languages, the Bastillefile that builds the jail plus vm-bhyve's .conf that configures the VM, and two lifecycles, bastille start for the jail and vm start for the guest inside it. That is where point 4 actually lives, and making the jail thinner does not close it.

If you drop vm-bhyve and have Bastille write the bhyve invocation directly into that thin hardened jail, you have closed 4, and you have also arrived at exactly what the branch already does. So we are converging on the same jail from two directions, and the only real question is whether Bastille drives bhyve or delegates to vm-bhyve inside. My branch is the former and it runs today.

On config staying confined to the jail: I like that property too, and the native version has it as well, just located differently. A VM is one instance dataset, vm.conf plus its zvols under vms/<name>, so moving it is a zfs send of that dataset. Yours is self-contained in vm-bhyve's world, its switches and its .conf; mine in Bastille's, one manifest with Bastille's own tap, bridge, and RDR conventions. Both are portable units, they just answer to different tools.

So real convergence on the jail. If there is appetite to try the experimental "Bastille writes the jail.conf and the bhyve args directly" option you mentioned in point 1, that is the native brand, and the branch is a working reference for it. Happy to line them up side by side on the same guest so we are comparing the actual thing, not the description of it.

tschettervictor:
Out of curiosity, have you looked at Appjail?

usenix17:
I had not, thanks for the pointer. I read through it.

As far as I can tell AppJail has no native bhyve support, the docs are all jail management. Where it does VMs it is the same pattern we have been discussing: a Makejail that stands up a jail with vm-bhyve installed inside (Overlord's ephemeral VMs do that, plus nbdkit).

So it is another example of delegating to vm-bhyve. And that is the point: none of the jail frameworks, Bastille, iocage, AppJail, drive bhyve natively. "Install vm-bhyve in a jail" is the common answer, and it is fine, but it is a recipe, not a brand.

Same fork as before: wrapping vm-bhyve is what AppJail already does, so yaazkal's scope question stands.

tschettervictor:
I don't think there's much point in just adding "vm jail" functionality and leaving it there.

It's either we go all the way with managing VMs with Bastille, or keep it to jails only, and let the user decide what to do with the jail from that point forward.

We can discuss more, but what do you think about creating a separate tool that does what you want, and possible uses Bastille for the initial creation of the jail?

Also, since one of your main concerns is a single management plane, what makes a supervised jail running a VM different from just a VM?

usenix17:
Agreed on no half-measures, and I think you have found the clean split. The full-drive version is what this branch already is, Bastille writes the jail.conf and the bhyve invocation and owns the VM as a first-class object. But if Bastille wants to stay jail-scoped, a separate tool that uses Bastille to create the jail and then drives bhyve inside it is a good shape, and I would rather build that than push a half-VM into Bastille. So your suggestion matches where I was already heading.

On the real question, what makes a supervised jail running a VM different from just a VM: nothing about the guest. It is the same bhyve, the same VM. The difference is entirely the container around the bhyve process. A bare bhyve on the host is a privileged process with the host's full /dev, the host network stack, and no rctl. The supervision jail wraps that one process in its own restricted devfs that exposes only that VM's nodes, its own vnet, and per-VM rctl. bhyve is a large C surface parsing untrusted guest I/O, so if it is compromised, an escape lands in an empty thin jail with almost no devices and no host network, instead of on the host. Same guest, much smaller blast radius, and the VM becomes a first-class object you can list beside jails and move with a zfs send of its dataset.

The honest version: if you do not care about confining bhyve, a supervised jail is close to just a VM plus bookkeeping. The reason to do it is the confinement and the single object model. That is also why the jail has to be thin and per-VM, one bhyve per jail, or you give the isolation back.

Happy to keep the branch as a working reference for whichever way this goes.

tschettervictor:
How does the VM work with a met jail as opposed to a shared IP or loopback jail? What IP does it get?

EDIT: i think I see how this is done going through the PR.

Another question arises however. What part of the VM is inside the jail? The disks are outside, the networking is outside, and bhyve is run from the host. Perhaps I'm missing something but it seems the jail becomes redundant at that point?

The bhyve command itself is run from the host, so doesn't that mean the VM is running on the host?

EDIT 2: I see now that it runs the vm.sh file in the root of the jail on startup.

tschettervictor:
I'd like to get your thoughts on #1600

usenix17:
Closing this out.

This started as an RFC for making a bhyve VM a first-class jail type inside Bastille core. The discussion made two things clear, and I agree with both: Bastille's core scope should stay small, and the VM lifecycle doesn't need to live in the main tree to work well with Bastille.

#1600 is what changes the calculus. With the plugin mechanism, the VM work can live in its own repo, install with `bastille -p <url>`, and run as `bastille -p vm ...` -- out of core, but still native to Bastille's UX. That lines up almost exactly with @tschettervictor's earlier point that the right shape is a separate tool that uses Bastille to create the jail. The plugin's `create.sh` calls `bastille create ...` and then brings up bhyve inside that jail. Bastille owns the jail (the confinement boundary: restricted devfs, own vnet, per-VM rctl); the plugin owns the VM.

So I'm going to build this out as a plugin against #1600 rather than push for a core merge. The branch on this PR already runs on bare metal, so it's a port of working code, not a rewrite. I'll keep the manifest/spec public so the design is reviewable, and I'll stay a consumer of the plugin API so any rough edges get reported back here (the `bootstrap_plugin` clone-target issue on #1600 is the first one).

Thanks @yaazkal and @tschettervictor for the back-and-forth -- it pushed this to a better home than where it started. Closing in favor of the plugin approach.

yaazkal:
thanks @usenix17 for the initiative

tschettervictor:
Please do share your plugin here. I'd like to see it in action...

PR #1600, opened 2026-07-25
Add support for plugins

tschettervictor:
Feat: Add support to use plugins

Steps:

1. Create a directory where your plugin will be stored. This directory MUST BE PLACED inside `bastille_sharedir/plugins`.
2. You must include the line `. /usr/local/share/bastille/common.sh` at the top of each file inside your plugin directory if you want your plugin to be able to use Bastille functions like `info` `warn` etc...
3. The plugin structure should be a top level directory named after your plugin, and any number of `CMD.sh` files inside it. See below

```
/usr/local/share/bastille/plugins/my-plugin/plugin-cmd1.sh
/usr/local/share/bastille/plugins/my-plugin/plugin-cmd2.sh
```

5. To run your plugin, use `bastille plugin PLUGIN_NAME PLUGIN_CMD` (You don't need to include .sh here)

See the following example to list bhyve VMs:

My plugin is at `/usr/local/share/bastille/plugins/vm-plugin` with one file inside called `list.sh`. (Must have .sh at the end)

The file has the following contents:

```
#!/bin/sh
list_bhyve() {
    vm list
}

list_bhyve
info 1 "Done!"
```

When I run `bastille plugin vm-plugin list` I now get:

```
NAME       DATASTORE  LOADER     CPU  MEMORY  VNC           AUTO     STATE
dev1       default    bhyveload  1    4G      -             No       Stopped
freepbx    default    uefi       2    8G      -             No       Stopped
zvault133  default    bhyveload  2    8G      -             No       Stopped
Done!
```

The plugin is basically just a collection of shell scripts, but this makes it easier to add functinality to bastille.


cedwards:
I think we're on the right track here and just need to think through some of the implementations. An exercise to think through possibilities.

What does this method look like if we try to integrate the DNS PR as a plugin?
What does this method look like if we try to integrate the vm-bhyve PR as a plugin?

tschettervictor:
I will try to do the DNS thing...

usenix17:
I like this a lot, and I think it's the right home for the bhyve-VM work in #1595.

We'd been going back and forth between two options for that PR: bake VM support into core, or ship it as a wholly separate tool that calls Bastille. This is a cleaner third path. A plugin lives in its own repo, installs with `bastille -p <url>`, and runs as `bastille -p vm ...`, so the VM management stays out of a core that's clearly leaning elsewhere while still presenting as first-class Bastille. That maps almost exactly onto your earlier suggestion of "a separate tool that uses Bastille to create the jail" -- the `create.sh` in the plugin just shells out to `bastille create ...` and then sets up bhyve inside that jail. Bastille owns the jail (the confinement boundary: restricted devfs, own vnet, per-VM rctl); the plugin owns the VM lifecycle. That's also the answer to your question on #1595 about the jail looking redundant -- the jail isn't the VM, it's the box the bhyve process runs in.

I'm happy to be the first real consumer and build the VM work out as a plugin against this.

One thing I hit reading the diff, before install-by-URL gets advertised:

```sh
git clone "${plugin}" "${bastille_sharedir}/plugins/"
```

This clones the repo contents straight into `plugins/` rather than into `plugins/<name>/`. Two problems: `git clone` refuses a non-empty target, so the second plugin anyone installs fails; and the layout won't match what `load_plugin` and the dispatch path expect (`plugins/${plugin}/${cmd}.sh`). Deriving the name from the URL fixes both:

```sh
plugin_name="${plugin##*/}"
plugin_name="${plugin_name%.git}"
git clone "${plugin}" "${bastille_sharedir}/plugins/${plugin_name}"
```

A couple of smaller things that would help once the basics are in: a version/compat check so a plugin can refuse to run against an incompatible Bastille, and some hook into `usage`/help so `bastille -p vm` without a subcommand lists what's available. Neither blocks merging.

Nice work -- this unblocks the VM direction without dragging it through core.

tschettervictor:
I've added a fix for the bootstrap and also added a hit pull if the plugin has already been installed. This should work as long as the plugin repo has not been touched. Otherwise you will have to run `git reset --hard` inside the plugin dir and run the bootstrap command again.

As for doing the help stuff, perhaps a help.sh file inside the plugin die would work. It all lies on the plugin to present relevant info/messages. Bastille will only call the files as they are on the command line.

yaazkal:
I wonder if we will need a "manifest" file for plugins. So the plugin creator can say information about the plugin like creator name, url, license, min Bastille version required and external dependencies like kernel modules to enable or any external pkg to install in the host. That way Bastille not only will be able to list plugin information based on the manifest but also automatically help install it checking its dependencies (optional)

usenix17:
Following up here: the bhyve VM work I mentioned building on this mechanism is now released as a standalone plugin.

**bastille-vm-plugin 1.0.0** -- https://github.com/usenix17/bastille-vm-plugin (release: https://github.com/usenix17/bastille-vm-plugin/releases/tag/v1.0.0)

It manages bhyve VMs as first-class peers of jails via `bastille -p vm ...`: each VM runs in an `allow.vmm` supervision jail with a restricted per-VM devfs and its own vnet. 1.0 covers the full lifecycle (create/start/stop/restart/console/list/clone/destroy), templates plus a template-less `--image` flag create, cloud-init (raw/qcow2 images), a framebuffer (auto-attached for RHEL-family GRUB), PCI passthrough, and cold migration (zfs send/recv between hosts). Validated end to end on FreeBSD 15 across Alpine/Rocky/Debian/Ubuntu, including a two-host migration.

So this PR's mechanism now has a real, sizable consumer, and it works well. Two bits of feedback from building against it:

1. `bootstrap_plugin` clones into `plugins/` rather than `plugins/<name>/` (the clone-target issue I flagged earlier), so install-by-URL needs the basename fix before it's advertised.
2. The bin consumes reserved option words (`version`, `help`, `-v`, ...) before plugin dispatch, so a plugin can't expose a `version`/`help` subcommand -- `bastille -p vm version` prints bastille's own version. Passing everything after `-p <plugin>` to the plugin verbatim would fix that.

Thanks for building this -- it's exactly the home this work needed.

tschettervictor:
Cool. One option would be to add support for a "plugin_name" file that could be pulled before cloning and a directory created for such.

Or how could we simplify the name to "vm" instead of the longer "bastille-vm-plugin"?

usenix17:
On the name: I'd fold it into the manifest @yaazkal floated -- a `name:` field is the cleanest way to get `bastille -p vm ...` from a repo called `bastille-vm-plugin`. The repo keeps its discoverable name; `bastille -p <url>` reads the manifest and clones into `plugins/${name}`. That beats a bare `plugin_name` file because the same manifest can carry the other things a real plugin needs.

Having just built a sizable one, every field I kept wanting maps to something I worked around by hand:

- `name` -- install dir + the `-p` word (the vm vs bastille-vm-plugin question).
- `min_bastille_version` -- the compat check I mentioned; a plugin can refuse to run against an incompatible Bastille instead of failing oddly.
- `depends` -- host prerequisites. The VM plugin has to `kldload vmm nmdm` itself on first use (jails can't autoload modules) and needs `sysutils/edk2-bhyve` (plus `qemu-tools` for qcow2). Something like:

  ```yaml
  name: vm
  min_bastille_version: "1.5"
  depends:
    kmod: [vmm, nmdm]
    pkg: [edk2-bhyve, qemu-tools]
  ```

  would let Bastille load the modules and check/install the pkgs (opt-in) instead of every plugin reinventing that.

Keep it optional: no manifest -> clone under the repo basename (today's behavior); with one, Bastille uses `name`, checks the version, and can offer to satisfy deps.

It also pairs with passing everything after `-p <plugin>` to the plugin verbatim (the reserved-word thing) -- then `version`/`help` become normal plugin subcommands that can read straight from the manifest.

Happy to be the guinea pig for a manifest format; the VM plugin exercises all of it (custom name, min-version, kmod + pkg deps), so it's a decent real-world test case.

tschettervictor:
I'd keep the manifest to a sysrc style.

name=name
min_version=1.5.0
depends_kmods="vmm nmdm"
depends_pkg="pkg1 pkg2"

Makes it way easier to parse for simple shell scripts.

usenix17:
Oh, good call -- sysrc-style is the right call for a shell tool, and it matches how Bastille already stores jail settings and how this plugin stores `vm.conf`. `sysrc -f <manifest> -n name` reads any field with zero deps, and the space-separated lists parse with a plain `for`.

Concretely, `bastille-vm-plugin` would ship (say `plugin.conf` at the repo root):

```sh
name="vm"
min_version="1.5.0"
depends_kmods="vmm nmdm"
depends_pkg="edk2-bhyve qemu-tools"
```

Bastille reads `name` before cloning (install into `plugins/${name}`), refuses if its own version is below `min_version` (a `sort -V` compare), and optionally `kldload`s the kmods / checks the pkgs. Everything else stays plugin-owned.

Happy to add exactly this to the VM plugin once there's a filename + field convention to target -- it's a real consumer that needs all four fields, so it's a good shakedown of the format.

tschettervictor:
Ok the PR has been updated.

One note, please name the pkg depend var as `depends_pkgs` with an s to stay consistent.

I was able to fully bootstrap your  plugin as well as update it. It's working nicely.

Also, make sure to pull bastille with git pull as there were some changes to the common.sh file

usenix17:
That's great to hear -- bootstrapped and updated cleanly is exactly the validation this needed. Done: renamed to `depends_pkgs` (plural) and pushed, so the repo's `plugin.conf` now reads:

```sh
name="vm"
min_version="1.5.0"
depends_kmods="vmm nmdm"
depends_pkgs="edk2-bhyve qemu-tools"
```

Will pull the common.sh changes, thanks for the heads-up. Out of curiosity -- did it land as `plugins/vm` off the `name` field, or still under the repo name for now? Either way, really nice to see install-by-URL + update working against a real plugin. Happy to keep exercising the manifest fields as they firm up.

tschettervictor:
It installs as the "name"

tschettervictor:
Also validated that it loaded and persisted the modules, as well as installed the packages.

Next is the documentation which I'm working on.

tschettervictor:
I think we will end up moving the plugin feature to a `plugin` subcommand or `p` for short. We are working on a global json output, and would prefer to use the `-p` for pretty, column output.

What do you think?

usenix17:
No strong preference here either, and if you want `-p` for pretty/column output, moving plugins to a `plugin` subcommand (with `p` as a short alias) is the clean answer. Reserving `-p` for a global pretty/JSON flag is a good use of it.

One bonus that actually tips me toward the subcommand: it fixes the reserved-word thing I hit earlier. As a top-level flag, `bastille -p vm version` gets eaten by the option parser (it prints Bastille's version). As a subcommand, `bastille plugin vm version` passes `version`/`help`/etc. straight through as the plugin's own CMD, so plugins can finally expose their own `version`/`help`. So the subcommand route buys you the `-p` flag and that fix.

From the plugin side it's a no-op, the dispatch changes but the `plugins/vm/*.sh` scripts don't, so I'll just update the docs to `bastille plugin vm ...` (or `bastille p vm ...`) once it lands. Happy either way, but the subcommand gets my vote.

tschettervictor:
Alternatively, we could do 'bastille vm create' where 'vm' is the plugin. I think there would be enough difference between the standard commands and plugin names.

And even if there is overlap, it would simply default to using the bastille command, which would then require you to prefix it with 'bastille plugin vm create'. But if there's no overlap, a shorter 'bastille vm' might be better.

tschettervictor:
PR updated.

You should now be able to bootstrap with `bastille plugin https:/....` and run it with either `bastille plugin vm...` or `bastille vm...`

Let me know your findings.

usr/local/share/bastille/plugin.sh
```diff
@@ -0,0 +1,157 @@
...
+            kldload -v "${kmod}" 2>/dev/null
+            info 1 "\nPersisting module: ${kmod}"
+            sysrc -f /boot/loader.conf ${kmod}_load=YES 2>/dev/null
+        done
+        # Install required plugin pkgs
+        for pkg in ${depends_pkgs}; do 
+            info 1 "\nInstalling package: ${pkg}"
+            pkg install -y ${pkg}
+        done
+        info 1 "Plugin bootstrapped. Use 'bastille -p|--plugin PLUGIN...' to run."
```
jdhg-orbiware: it says to use -p|--plugin when those flags actually does not exist here

docs/index.rst
```diff
@@ -17,24 +17,17 @@ https://docs.bastillebsd.org.
    :maxdepth: 2
    :caption: Contents:
 
-   chapters/comparing
-   chapters/installation
-   chapters/getting-started
-   chapters/configuration
-   chapters/targeting
+   chapters/overview/index
```
jdhg-orbiware, line 20: full rewrite of docs deserve another PR, let's try to keep this PR clean and only include the plugins docs

usr/local/bin/bastille
```diff
@@ -99,6 +99,7 @@ Available Commands:
   mount       Mount file(s)/directorie(s) inside jail(s).
   network     Add or remove interface(s) from jail(s).
   pkg         Manage packages inside jail(s). See pkg(8).
+  plugin      Run a custom Bastille plugin.
```
jdhg-orbiware, line 108: For better control, as in templates, this will need 'bastille_pluginsdir="${bastille_prefix}/plugins" in the Bastille's configuration file

usr/local/share/bastille/plugin.sh
```diff
@@ -0,0 +1,157 @@
...
+        # Clone plugin repo
+        if ! git clone "${plugin_url}" "${bastille_sharedir}/plugins/${name}"; then
+            error_exit "[ERROR]: Failed to bootstrap plugin."
+        else
+            info 1 "Plugin bootstrapped. Use 'bastille -p|--plugin PLUGIN...' to run."
+        fi
+        # Load required plugin modules
+        for kmod in ${depends_kmods}; do 
+            info 1 "\nLoading module: ${kmod}"
+            kldload -v "${kmod}" 2>/dev/null
```
jdhg-orbiware: can we catch if there was any error here? that way we prevent writing bad things in the next line. Specially because we are going to touch a pretty sensitive host file.

usr/local/share/bastille/plugin.sh
```diff
@@ -0,0 +1,157 @@
...
+            info 1 "\nPersisting module: ${kmod}"
+            sysrc -f /boot/loader.conf ${kmod}_load=YES 2>/dev/null
+        done
+        # Install required plugin pkgs
+        for pkg in ${depends_pkgs}; do 
+            info 1 "\nInstalling package: ${pkg}"
+            pkg install -y ${pkg}
```
jdhg-orbiware: as in kldload... should we need to catch errors here? If it fails should we actually install the plugin? My take is if kldload or if pkg fails we should not install the plugin.

usr/local/share/bastille/lib/plugin.sh
```diff
@@ -0,0 +1,159 @@
...
+    if [ -d "${bastille_sharedir}/plugins/${name}" ]; then
+        if ! git -C "${bastille_sharedir}/plugins/${name}" pull; then
+            error_exit "[ERROR]: Failed to update plugin."
+        fi
+    else
+        # Clone plugin repo
+        if ! git clone "${plugin_url}" "${bastille_sharedir}/plugins/${name}"; then
+            error_exit "[ERROR]: Failed to bootstrap plugin."
+        else
+            info 1 "Plugin bootstrapped. Use 'bastille -p|--plugin PLUGIN...' to run."
```
jdhg-orbiware, line 154: -p|--plugin does not exist in practice

jdhg-orbiware (changes requested):

usr/local/share/bastille/plugin.sh
```diff
@@ -0,0 +1,159 @@
...
+shift 2
+
+bastille_root_check
+
+bootstrap_plugin() {
+
+    local plugin_url="${1}"
+    local plugin_name="$(basename "${plugin_url}" | sed 's/.git//')"
+
+    if echo "${plugin_url}" | grep -q "github.com"; then
```
jdhg-orbiware: This will actually support whatever has github.com on his name. Like 'evilgithub.com' will pass

usr/local/share/bastille/lib/plugin.sh
```diff
@@ -0,0 +1,159 @@
...
+bastille_root_check
+
+bootstrap_plugin() {
+
+    local plugin_url="${1}"
+    local plugin_name="$(basename "${plugin_url}" | sed 's/.git//')"
+
+    if echo "${plugin_url}" | grep -q "github.com"; then
+        local repo="$(echo "${plugin_url}" | awk -F"github.com/" '{print $2}' | sed 's/.git//')"
+        local manifest_url="https://raw.githubusercontent.com/${repo}/main/plugin.conf"
```
jdhg-orbiware, line 116: I guess cloning in a temp dir and do this instead of pointing to this domain can open the door to support other git services. Is this because only github is supported?

usr/local/share/bastille/lib/plugin.sh
```diff
@@ -0,0 +1,159 @@
...
+    else
+        error_exit "[ERROR]: Only supports github at this time."
+    fi
+    # Check for manifest file
+    local manifest="$(mktemp)"
+    if ! fetch -o "${manifest}" "${manifest_url}"; then
+        warn 1 "[WARNING]: No 'plugin.conf' found. Using repo name as plugin name."
+        local name="${plugin_name}"
+    else
+        local name="$(sysrc -f "${manifest}" -n name 2>/dev/null)"
```
jdhg-orbiware, line 126: if the name is empty in the plugin config file we are not catching it here

usr/local/share/bastille/lib/plugin.sh
tschettervictor: Cloning would mean retrieving the entire repo. If in the future we want to support more, we can do it sure.

docs/index.rst
tschettervictor: Oops, did I push these here. I will use another

usr/local/share/bastille/lib/plugin.sh
```diff
@@ -0,0 +1,159 @@
...
+        local min_version="$(sysrc -f "${manifest}" -n min_version 2>/dev/null)"
+        local depends_kmods="$(sysrc -f "${manifest}" -n depends_kmods 2>/dev/null)"
+        local depends_pkgs="$(sysrc -f "${manifest}" -n depends_pkgs 2>/dev/null)"
+        # Validate plugin version against Bastille version
+        if [ "$(echo "${min_version}" | sed 's/\.//g')" -gt "$(bastille version | sed 's/\.//g')" ]; then
```
jdhg-orbiware: if min_version is empty in config file I guess this will fail. Also note that if min_version is "2.0" but current version is "1.4.4" will also fail, not to mention if the dev wants to add suffixes like -rc -beta -alpha or whatever.

jdhg-orbiware (changes requested):
adding more comments

usr/local/share/bastille/lib/plugin.sh
```diff
@@ -0,0 +1,167 @@
...
+
+case "${PLUGIN}" in
+    http?://*/*/*)
+        bootstrap_plugin "${PLUGIN}"
+        exit 0
+        ;;
+    *)
+        if [ -z "${PLUGIN_CMD}" ]; then
+            usage
+        elif ! echo "${PLUGIN_CMD}" | grep -Eoq '^[a-zA-Z0-9-_]+$'; then
```
jdhg-orbiware: This is rejecting every plugin subcommand. This regex should do the trick: '^[A-Za-z0-9_-]+$'

usr/local/share/bastille/lib/plugin.sh
```diff
@@ -0,0 +1,167 @@
...
+        if ! git clone "${plugin_url}" "${bastille_sharedir}/plugins/${name}"; then
+            error_exit "[ERROR]: Failed to bootstrap plugin."
+        else
+            info 1 "Plugin bootstrapped. Use 'bastille -p|--plugin PLUGIN...' to run."
+        fi
+        # Load required plugin modules
+        for kmod in ${depends_kmods}; do 
+            info 1 "\nLoading module: ${kmod}"
+            if ! kldload -v "${kmod}" 2>/dev/null; then
+                error_continue "[ERROR]: Failed to load module: ${kmod}"
```
jdhg-orbiware: we should stop the installation of the plugin, this will show at the end that the plugin was bootstraped, same for the error in pkg

jdhg-orbiware (changes requested):

usr/local/share/bastille/lib/plugin.sh
tschettervictor: I adjusted this to work for numbers only. Not really sure how we would do dev branches as they are names after the commit. I think just defaulting to not running the check is ok for now

usr/local/bin/bastille
tschettervictor: Lets keep in share_dir for now.

usr/local/share/bastille/plugin.sh
```diff
@@ -0,0 +1,171 @@
...
+shift 2
+
+bastille_root_check
+
+bootstrap_plugin() {
+
+    local plugin_url="${1}"
+    local plugin_name="$(basename "${plugin_url}" | sed 's/.git//')"
+
+    if echo "${plugin_url}" | grep -q "https://github.com"; then
```
jdhg-orbiware: puttin`https://` here will discard any repo that has a url like `git@github.com:user/repo.git`

This url format is common for clonning using ssh instead of https, useful for private repos. Templates support them.

jdhg-orbiware (changes requested):

usr/local/share/bastille/plugin.sh
tschettervictor: They do, but plugins don't right now. But it is a good idea. Will do that...

jdhg-orbiware (approved):

PR #1610, opened 2026-07-29
Support json (and pretty) output

jdhg-orbiware:
This is a draft and only `config get` is supported as a test case scenario. Opening this draft to discuss the design decisions and get feedback. The intention here is to be more API friendly and do some exploration of this capabilities.

Many FreeBSD base tools support machine-readable output via libxo's --libxo=json (jls, netstat, ps, arp, mount, vmstat, …). Bastille today has only one JSON output at the moment: `bastille list -j`, which is hand-rolled in `list.sh` with `awk` and a `sed` pretty-printer.

Every libxo JSON document is a single top-level object shaped like:

`{ ["__version": "N",]  "<domain>": { "<singular-element>": [ {record}, {record}, … ] } }`

The idea here for Bastille is to stay as close as possible to that format. Nevertheless, I find adding a "type" because some commands can be executed at different target-types. Like `bootstrap` subcommand can do releases or templates. This can be discussed if really needed or not depending if we are going to support a json output for `bootstrap`.

So the Bastille output is something like:

`{"bastille":{"type":"<entity>","<entity>":[ {record}, ... ]}}`

## How to test this

For json output:
`# bastille -j config TARGET get securelevel`

For json (pretty) output:

`# bastille -p config TARGET get securelevel`

`ALL` is supported as target.

Both print all jails with its jid, name, and securelevel or whatever config you are testing.

If the jail is not running, the `jid` will simply say `null`


jdhg-orbiware:
Rebase to use the new 'lib' directory

tschettervictor:
Works for me. Good to see the new LIBDIR.

Perhaps it's better to use "${bastille_sharedir}/lib/json.sh" so it's dynamic.

jdhg-orbiware:
> Perhaps it's better to use "${bastille_sharedir}/lib/json.sh" so it's dynamic.

You're right. I'll also created another PR #1614 in order to update the other scripts in the common.sh file.

I'll wait @cedwards and @bmac2 comments before continuing adding other commands support. Mainly to be sure the design is good enough so I can start adding more code.

tschettervictor:
I'd like to start with merging this as a first. I've got an idea about moving supported parameters or config into a lib/config file so we don't have to pollute the config.sh subcommand.

But let's wait until bmac and cedwards ok the format here.

jdhg-orbiware:
I've made a rebase and ready to merge. Please review/aprobe so I can continue.

usr/local/bin/bastille
```diff
@@ -121,6 +121,8 @@ Use "bastille COMMAND -h|--help" for more information about a command.
 Use "bastille -c|--config FILE COMMAND" to specify a non-default config file.
 Use "bastille -v|--version" for version information.
 Use "bastille -q|--quiet COMMAND" to supress Bastille-specific output.
+Use "bastille -j|--json COMMAND" for JSON output (on supported commands).
+Use "bastille -p|--pretty COMMAND" for indented JSON output (implies --json).
```
tschettervictor: I think we should change this and use -p for plugins. -c is also used so we can't use that. What else is there?

usr/local/share/bastille/config.sh
```diff
@@ -169,6 +169,9 @@ print_jail_conf() {
 '
 }
 
+# Emit JSON only for the 'get' action (a query); other actions are unchanged.
+[ "${ACTION}" = "get" ] && json_open jail
```
tschettervictor: Should we check for the global json flag before running this?

tschettervictor (approved):

PR #1615, opened 2026-07-30
Extending tests with shellspec unit suite

cedwards:
This introduces shellspec unit test suites that allow us to better validate function behavior.

These changes have been integrated with the new lib/ structure and call specific lib files to test only those bits.

75 tests have been defined and validated. More are pending following some function re-structuring (to be discussed separately).


cedwards:
I assume the warning tests are due to executing on GitHub actions Ubuntu environment.

They all pass when I `make test` on my macOS and FreeBSD environments.

Confirmed: it's the difference between BSD sed and GNU sed. Fixing.

.github/workflows/freebsd-e2e.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Run Bastille test suite in FreeBSD
+        uses: vmactions/freebsd-vm@v1
+        with:
+          usesh: true
+          prepare: |
+            pkg install -y git
```
jdhg-orbiware, line 31: git-lite will be faster here, unless there is any reason to have the bloated git

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,31 @@
+name: Unit Tests
+
+# Fast, host-agnostic shellspec unit tests. These gate every PR: they run in
```
jdhg-orbiware, line 3: We already have shellcheck host-agnostic here in the repo, how is this different than the other? Also the other have some options configured. Maybe this is different and needed for the tests unit?

cedwards: They do different things.

Shellcheck is a static analysis tool for shell scripts.
Shellspec is a full-featured BDD unit testing framework.

.github/workflows/freebsd-e2e.yml
cedwards: Sure, I can change this.

tschettervictor:
My only comment would be to first go through the duplicated functions and try to merge them, and also decide on a standard naming convention for functions.

.github/workflows/freebsd-e2e.yml
tschettervictor: Maybe even git-tiny

jdhg-orbiware: @cedwards I'll like to review/aprove this. Do you want to keep git or want me to change it?

jdhg-orbiware (changes requested):

.github/workflows/freebsd-e2e.yml
```diff
@@ -0,0 +1,50 @@
...
+            # orchestrator Bastillefiles (tests/ufs-tests/init). They drive
+            # host-level commands (setup/bootstrap/create/...) rather than
+            # targeting an existing jail, so the template must be resolvable
+            # from bastille_templatesdir or invoked by full path. Wire the
+            # confirmed command in below, e.g.:
+            #
+            #   cp -a tests "$(bastille config -g bastille_templatesdir)"/
+            #   bastille template ALL ufs-tests/init
+            #
+            echo "TODO: invoke tests/ufs-tests/init once the runner command is confirmed"
```
jdhg-orbiware, line 50: This will give us always green, is not running any tests, just echoing the message. Maybe we can merge the unit tests and this tier 2 tests later?

PR #1619, opened 2026-08-04
Add POC for OCI jails

tschettervictor:
FEAT: Support for creating a jail based entirely on an OCI image. This is POC and experimental, so give it a go.

To get started, verify you have the following variable set in `/usr/local/etc/bastille.conf`: `bastille_volumesdir="${bastille_prefix}/volumes"`

To create an OCI jail, use the new `-O|--oci` flag when creating a new jail. Additional options include:

* `-e KEY=VALUE` to pass environment variables to the jail (repeatable)
* `--data-path PATH` to specify a location for persistent data
* `--os OS` to pull non-freebsd images (eg: `--os Linux`)

Bastille extracts most information from the image itself, and stores important info inside the jail conf directory. This includes things like volume mounts, entrypoint, labels, env vars, start cmd, stop signal, and a few more. Because it knows the volumes it needs to mount, you need not worry about persistent data, unless you want to store it somewhere other that `${bastille_volumesdir}`

## Creating an OCI jail using the create sub-command

To create an OCI jail using the navidrome image from daemonless.io, run the following command:
```
bastille create -O navidrome ghcr.io/daemonless/navidrome 10.12.12.12
```
Notice that instead of the RELEASE, we simply specify an image name. This will fetch the image, extract it into the jail, and configure jail.conf to launch the entrypoint on startup.

The image pulled will always be latest. For now, if you want to update an image, simply destroy the jail and rebuild it. All persistent data is stored either in your custom directory (if `--data-path` was given at create) or at `${bastille_volumesdir}`

It is possible to create a VNET jail using this method with either `-B` or `-V` and the app should run fine. Just note that you do not have to forward any ports for VNET jails.

It is also possible to pull linux images using the `--os linux` flag. Tested with docker.io/filebrowser/filebrowser but don't expect support for them, as even though some will work, they are not supported on FreeBSD.

The following examples all work in testing:
```
bastille create -O navidrome ghcr.io/daemonless/navidrome 10.12.12.12
bastille create -O --data-path /usr/local/apps filebrowser ghcr.io/appjail-makejails/filebrowser 10.12.12.13
bastille create -O --os linux filebrowser-linux docker.io/filebrowser/filebrowser 10.12.12.14
bastille create -O -e OC_URL=https://192.168.1.10:9200 -e OC_INSECURE=true opencloud ghcr.io/daemonless/opencloud 10.12.12.15
```
Notice that for opencloud you need to specify your host IP as the OC_URL instead of the jail IP. Opencloud in finicky in this way.

## Creating an OCI jail using the up sub-command

In addition to these, a new sub-command has been added. `bastille up` will attempt to deploy an OCI image from a `podman-compose.yml` file in the current working directory.

Some things to note for `bastille up`:
* Bastille will always run `bastille up` from the $PWD, so make sure you are in the directory that has the compose file
* Bastille doesn't care about volumes. It stores all data inside `${bastille_prefix}/volumes/${JAIL}` or `--data-path` so you don't have to worry about that
* Bastille will set the IP to `inherit` if podman-compose has `network_mode: host`. Change to `network_mode: 10.34.34.34` to use that specific IP for that specific container
* Bastille will auto forward ports for you if they are given in the image and the jail is a NAT jail
* You can also use `--data-path` in `bastille up` to specify a non-default path. (Don't include the jail name here, only the top level path like `/usr/local/apps`
* Bastille will look for a `.env` file in the directory where the podman-compose file is, and use the values inside it to replace any instances of a given variable in the compose file. See the following example:

Taken from https://github.com/daemonless/audiobookshelf

podman-compose.yml
```
services:
  audiobookshelf:
    image: "ghcr.io/daemonless/audiobookshelf:latest"
    container_name: audiobookshelf
    environment:
      - PUID=${PUID}  # User ID for the application process
      - PGID=${PGID}  # Group ID for the application process
      - TZ=${TZ}  # Timezone for the container
    volumes:
      - "/path/to/containers/audiobookshelf:/config"
      - "/path/to/containers/audiobookshelf/metadata:/metadata"
      - "/path/to/containers/audiobookshelf/audiobooks:/audiobooks"
    ports:
      - "13378:13378"
    restart: unless-stopped
```
.env
```
PUID=1000
PGID=1000
TZ=America/Edmonton
```
Simply enter the directory where these file are located and run `bastille up`

Not all annotations of the compose file are supported. For example, Bastille stores all persistent data a `${bastille_volumesdir}` regardless of the `volumes:` annotation in the compose file. Most options should be supported, but not all are used. The parser will create a `bastille-compose.conf` file next to the `podman-compose.yml` file in the following format. Fore each indentation, we simply add one level. For example, `environment:` in the example above will read:
```
audiobookshelf_environment=PUID:1000
audiobookshelf_environment=GUID:1000
audiobookshelf_environment=TZ:America/Edmonton
```
We use `:` to better handle env vars when building the create command that will be run at the end of the `bastille up` routine.

## Notes

This is still experimental, and code standardization will need to be figured out. This includes the jails structure, as well as the jail.conf file and any additional flags to add.


cedwards:
I've done some basic testing against daemonless.io containers and everything works as expected. Feature looks good. Great job on this!

tschettervictor:
I'm some documentation and merge it after some last minute testing. If nothing is amiss that is...

PR #1630, opened 2026-08-20
config: Supports multiple comma separated get values

jdhg-orbiware:
Currently you can only get one only value from the jail config, like `bastille config alcatraz get securelevel` and that will print the value. But you hit a limitation if you want to get multiple config values. This PR adds that possibilty:

```
ishmael ~ # bastille config alcatraz get name,securelevel,osrelease
  PROPERTY     VALUE
  name         alcatraz
  securelevel  2
  osrelease    15.1-RELEASE
```

Currently if you want to print a property using ALL as target, you just get a list of values, but you have no idea what value is to what jail. 

This PR allows getting a property from every jail. Multiple jails always print a `JID JAIL PROPERTY VALUE` table (stopped jails show `-` in the `JID` column):

```
  ishmael ~ # bastille config ALL get securelevel
  JID  JAIL         PROPERTY     VALUE
  2    alcatraz     securelevel  2
  3    bella        securelevel  2
  1    gorgona      securelevel  2

```

Several properties across every jail use the same four columns, one row per jail and property:

```
  ishmael ~ # bastille config ALL get name,securelevel
  JID  JAIL         PROPERTY     VALUE
  2    alcatraz     name         alcatraz
  2    alcatraz     securelevel  2
  3    bella        name         bella
  3    bella        securelevel  2
  1    gorgona      name         gorgona
  1    gorgona      securelevel  2
```

This PR makes sure JSON support is not broken and supported in every case.

### DIFF
--- a/docs/chapters/subcommands/config.rst
+++ b/docs/chapters/subcommands/config.rst
@@ -15,6 +15,17 @@ Getting a property that *is not* defined in jail.conf
   ishmael ~ # bastille config azkaban get notaproperty
   not set
 
+Getting several properties at once (comma-separated). A single property
+still prints only its value; multiple properties print a table:
+
+.. code-block:: shell
+
+  ishmael ~ # bastille config alcatraz get name,securelevel,osrelease
+  PROPERTY     VALUE
+  name         alcatraz
+  securelevel  2
+  osrelease    15.1-RELEASE
+
 Setting a property:
 
 .. code-block:: shell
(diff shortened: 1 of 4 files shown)


tschettervictor:
I think a more Bastille like approach would be to print like so:

```

[jail1]:
securelevel=2
osrelease=15.1-RELEASE

[jail2]:
securelevel=2
osrelease=15.1-RELEASE
```

Or even

```

[jail1]:
securelevel 2
osrelease 15.1-RELEASE

[jail2]:
securelevel 2
osrelease 15.1-RELEASE
```

It keeps the output more consistent.

jdhg-orbiware:
You are right. I took 'zfs get' as an inspiration, thinking that a table format (like the list subcommand does it) is more scriptable, what do you think? do you want me to change this for the format you are saying anyway?

tschettervictor:
I would say doing 'securelevel=2' is more scriptable. That way users can do the config command, then awk -F= '{print $2}'

PR #1662, opened 2026-09-04
bastille-update: Update child jail osrelease on proper conditions

linnemannr:
'bastille update' incorrectly updates the osrelease of any lightweight jail, regardless of which release jail it references in its fstab. Change the match regex to match the release jail dir as the device in the fstab entry to ensure only lightweight jails that explicitly derive from the updated release are given the new osrelease.


tschettervictor:
Thanks!

linnemannr:
Thank you for taking the PR!

